PR #39, opened 2012-04-03
RPC Client now has ability to be created in prototype scope

o10g:
Hello,

when I started to integrate your Bundle to my application I realized, that I need to create more then one instance for one RPC client.

If you would like, I could demonstrate why I have to have different instances for RPC Clients.

Thank you,
Oleg


Enleur:
@maectpo scopes now are deprecated in symfony and removed in 3.x branches

PR #130, opened 2013-09-03
[RPC] Add route option for rpc-server command

tyx:
Allows to run rpc server for a route specified


videlalvaro:
Do you think you could document this new behaviour as well?

PR #158, opened 2014-01-28
Add interaction between consumer and his callback

metfan:
I'm having some troubles with long time running consumer, I loose MysQL connection and can't reopen it.
In this case I need to requeue my message and stop the consumer to start a fresh one.

The only way I found to stop consumer is to throw an exception but in this case I can't return ConsumerInterface::MSG_SINGLE_NACK_REQUEUE at the end of execute() method in my callback.

In BaseConsumer class there is code to stop consumer and I don't want to inject consumer in my callback (circle reference, not good).

So I propose some change:
- don't inject a callable in BaseConsumer::setCallBack() but the real service and type hint the parameter
- replace use of call_user_func() by calling the execute() method
- extend ConsumerInterface with a new InteractiveConsumerInterface
- test use of InteractiveConsumerInterface at the end Consumer::processMessage() to potentially stop consume

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,5 @@
+- 2014-01-28
+    * adds new interface InteractiveConsumerInterface to allow communication between consumer and his callback. at this time, only implements a way to stop consume from callback.
 - 2013-01-18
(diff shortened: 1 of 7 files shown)


videlalvaro:
closing since now we have #230

PR #226, opened 2014-12-11
Fixed handling of system signals when waiting for a new message from the...

### LINKED ISSUE #178: Signals not handled when the queue is empty
When I try to send a signal to a consumer with an empty queue, it is not taken into account until a message arrives and is completely processed. I tried with `SIGINT` and `SIGTERM`. I'm using v1.3.2, and running the `rabbitmq:consumer` command without any flag (especially not `-w` because I don't want to interrupt a consumer that processes a message).

TwistedLogic:
... queue. Close #178

### DIFF
--- a/Command/BaseConsumerCommand.php
+++ b/Command/BaseConsumerCommand.php
@@ -3,6 +3,7 @@
 namespace OldSound\RabbitMqBundle\Command;
 
 use OldSound\RabbitMqBundle\RabbitMq\BaseConsumer as Consumer;
+use PhpAmqpLib\Exception\AMQPTimeoutException;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


videlalvaro:
Can you explain the fix?

TwistedLogic:
If you launch a consumer without the -w option, it's assumed the consumer will receive and process OS signals (SIGTERM, SIGINT, SIGHUP). If a message is currently to be processed everything works well, the message is processed and the consumer quits. When the consumer is waiting for new mesages, i.e. the queue is empty, if you send a SIGINT signal for instance, nothing happens because the BaseConsumerCommand::forceStopConsumer() method just sets the protected property $forceStop and does nothing else. The BaseConsumer::stopConsuming() method instead stops the wait on the socket, so the consumer can use the exit signal even when the queue is empty.

I'm not completely sure if there could be a message loss, can you take a look at it?

RuslanZavacky:
@videlalvaro @TwistedLogic any news on merging it?

PR #227, opened 2014-12-12
exit from stalled consumers

### LINKED ISSUE #180: Provide a way to reject message and exit consumer
I have a Consumer that uses Doctrine2. When a doctrine error occurs, the entity manager gets closed and usable. Rebuilding this object is very cumbersome. The simplest way to move on would be to simply restart the consumer. Is there a way to achieve this ? For the moment, I do this at the top of my script:

```php
if (!$this->em->isOpen()) {
     $this->logger->err('The entity manager is closed. Exiting.');
     exit(1);
 }
```

By doing this, the consumer exits when receiving a (probably fine) message *after* the message that caused the error, supervisord restarts it because it sees the status is not 0 or 2 (supervisord default `exitcodes` setting value), and the fine message is processed again. All is well, but I think there could be a `ConsumerInterface` constant to state "reject message and exit with 1". That would be cleaner.

bburnichon:
closes #180


bburnichon:
Closing this as #230 was merged instead

PR #236, opened 2015-01-27
Support for route and format for StdInProducer

jaymecd:
This PR provides following:

- [x] support for routing key

        $ echo -n 'data' | ./app/console rabbitmq:stdin-producer -r 'my.key' my_producer

- [x] support for payload format

    PHP serialization, default behavior:

        $ echo -n '{"name":"Germany","cnt":1234}' | ./app/console rabbitmq:stdin-producer my_producer
        # puts into queue: string(37) "s:29:"{"name":"testing","cnt":1234}";"

    Raw data, pre-formatted:

        $ echo -n '{"name":"Germany","cnt":1234}' | ./app/console rabbitmq:stdin-producer my_producer -f raw
        # puts into queue: string(29) "{"name":"testing","cnt":1234}"
    
- [ ] update Readme

### DIFF
--- a/Command/StdInProducerCommand.php
+++ b/Command/StdInProducerCommand.php
@@ -9,7 +9,9 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class StdInProducerCommand extends BaseRabbitMqCommand
 {
-
+    const FORMAT_PHP = 'php';
+    const FORMAT_RAW = 'raw';
+    
     protected function configure()
     {
         parent::configure();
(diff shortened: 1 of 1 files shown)


videlalvaro:
I think this solution is too specific. It would be better to have a way to specify a MessageEncoder/Decoder and then be able to use that from the producer/consumer.

Also AMQP already has `content-type` fields for messages, and this seems like an opportunity to properly use that.

See http://www.enterpriseintegrationpatterns.com/FormatIndicator.html

jaymecd:
Agree, but std-in producer is also pretty specific, mostly for debug.
The problem I saw recently, is application uses JSON payload, but std-in producer uses PHP serialized payload only. Mean I have to write own std-in produces, but I'm not sure it worth it.

videlalvaro:
@jaymecd ok, since this affects only the StdIn producer I'll merge it, thanks!

jaymecd:
:tada: Thank you, but a bit too fast: wanted to update readme as well.

Command/StdInProducerCommand.php
```diff
@@ -39,7 +43,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        switch ($format) {
+            case self::FORMAT_RAW:
+                break; // data as is
+            case self::FORMAT_PHP:
+                $data = serialize($data);
+                break;
+            default:
+                throw new \InvalidArgumentException(sprintf('Invalid payload format "%s", expecting one of: %s.',
+                    $format, implode(', ', [self::FORMAT_PHP, self::FORMAT_RAW])));
```
j0k3r: This won't work in PHP < 5.4 and the bundle should be compatible for PHP >= 5.3.0, according to the `composer.json`.

jaymecd:
@j0k3r, nice catch. Tnx

PR #237, opened 2015-01-28
Review Symfony components dependencies

ch3ric:
HttpKernel and FrameworkBundle are not necessary for a simple console application.
This bundle could be used in a console application, with Symfony Console and Dependency Injection components, to run RabbitMQ consumers.

So I changed the dependendies in composer.json from "symfony/framework-bundle" to "symfony/dependency-injection" and other mininum required Symfony components.

I also changed the minimum supported version for Symfony from 2.0 to 2.3 (Symfony < 2.3 is not supported by SensioLabs anymore)

### DIFF
--- a/Command/BaseRabbitMqCommand.php
+++ b/Command/BaseRabbitMqCommand.php
@@ -2,8 +2,30 @@
 
 namespace OldSound\RabbitMqBundle\Command;
 
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand as Command;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Console\Command\Command;
 
-abstract class BaseRabbitMqCommand extends Command
+abstract class BaseRabbitMqCommand extends Command implements ContainerAwareInterface
 {
+    /**
+     * @var ContainerInterface
+     */
+    protected $container;
+
+    /**
+     * {@inheritDoc}
+     */
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @return ContainerInterface
+     */
+    public function getContainer()
+    {
+        return $this->container;
+    }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ matrix:
         - php: hhvm
 
 env:
-    - SYMFONY_VERSION="2.1.*"
-    - SYMFONY_VERSION="2.2.*"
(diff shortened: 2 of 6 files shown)


lepiaf:
:+1:

ClementGautier:
We could do better by splitting this bundle in a component + bundle. This way the bundle could still depends on framework-bundle and the component only on the required libraries. But, as it can be time consuming I guess I give my :+1:  on this one too.

ch3ric:
ping @videlalvaro :)
What do you think about this PR? Could it be merged?

videlalvaro:
@ch3ric how much will this change break BC?

ch3ric:
Hello @videlalvaro,
I think the only thing it breaks is that old Symfony versions are not supported anymore, due to the constraints added in the file composer.json, and `Symfony\Component\DependencyInjection\Extension\Extension` has been added in the component dependency-injection since version 2.2, for example.
That's why I recommend to create a new version 1.6 after this PR will be merged.

videlalvaro:
I will merge a couple of PRs I see, do some patch, releases and then bump again the minor version number with your changes

ch3ric:
Thanks @videlalvaro 
I just rebased my branch.

videlalvaro:
Thanks!

Command/BaseRabbitMqCommand.php
```diff
@@ -2,8 +2,30 @@
 
 namespace OldSound\RabbitMqBundle\Command;
 
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand as Command;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Console\Command\Command;
 
-abstract class BaseRabbitMqCommand extends Command
+abstract class BaseRabbitMqCommand extends Command implements ContainerAwareInterface
```
vadim2404: For what this class has been created? Will you use a ContainerAwareCommand instead of this one?

ch3ric:
thanks @videlalvaro

Command/BaseRabbitMqCommand.php
ch3ric: ContainerAwareCommand is in the Symfony FrameworkBundle: we don't want to depend on this component (in case this bundle is used in a console command that does not use Symfony fullstack framework)

ch3ric: @vadim2404 are you facing an issue because of this change?

vadim2404: No. I searched only an answer for your solution. Thanks for explanation @ch3ric

Issue #239, opened 2015-02-16
Update composer.json to use amqplib 2.4.1


Intrepidity:
2.4.0 keeps throwing segfaults as was shown in issue 183 on that bundle. Should be solved in 2.4.1, but with the current version of RabbitMqBundle my composer does not want to update it.

videlalvaro:
I don't quite understand the question, but feel free to submit a PR

PR #240, opened 2015-02-16
Fixes issue #239

Intrepidity:
As 2.4.0 of the amqplib throws segfaults, I changed the RabbitMqBundle to accept newer minor versions of 2.4. This should solve these issues.


Intrepidity:
Never mind. I see the issue is still present in 2.4.1

PR #245, opened 2015-03-09
fix consumed counter to work on consecutive calls

### LINKED ISSUE #244: Reset consumed property when calling consume().
We have the problem that we call multiple times the consume() method with parameter 1 in the Consumer class. Only the first call works as expected because the consumed property isnt reset but always considered in the maybeStopConsumer method.

I will work on a PR for that.

marco-jantke:
fix #244 

didnt know how to test easily :/

### DIFF
--- a/RabbitMq/BaseConsumer.php
+++ b/RabbitMq/BaseConsumer.php
@@ -97,4 +97,13 @@ abstract class BaseConsumer extends BaseAmqp implements DequeuerInterface
     {
         return $this->idleTimeout;
     }
+
+    /**
+     * Resets the consumed property.
+     * Use when you want to call start() or consume() multiple times.
+     */
+    public function resetConsumed()
+    {
+        $this->consumed = 0;
+    }
 }


videlalvaro:
So the idea is to reuse a consumer instance, therefore the `consumer` variable needs to be re initialised ?

marco-jantke:
Yes, we are calling the consume method multiple times in one process. 
Our usecase: we have integration tests where we put messages to a queue and consume them stepwise, asserting a certain state in our system and then consume the next one.

videlalvaro:
Wouldn't it be better to add a resetConsumed() method to prevent BC breaks?

marco-jantke:
For me this seems like expected behavior so I would have fixed it. Tough I see the point of BC breaks.. So its up to you to decide ;-)

videlalvaro:
I think at the beginning it was not thought to call start consumer multiple
times, so I would prefer an extra method call

PR #249, opened 2015-03-17
Switched to PSR-4 instead of using target-dir in composer

mhor:


videlalvaro:
is there any BC break with this approach?

mhor:
There is no BC break with this approach.

Issue #148, opened 2013-12-17
Sandbox implementation is flawed


stof:
Currently, turning on the sandbox mode in the config (which seems meaningful for a test environment) is likely to break the code with a Fatal error (it would in my app for instance).

The behavior of this option is to switch producer services from the [Producer](https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/Producer.php) class to the [Fallback](https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/Fallback.php).
But the Fallback class does not extend Producer, which means it will not pass the typehints

mbiagetti:
:+1:

videlalvaro:
can this issue be closed now?

PR #251, opened 2015-03-26
add ProducerInterface

agiuliano:
In sandbox mode, instead of `Producer`, `old_sound_rabbit_mq.my_name_producer` is mapped on the `Fallback` class.
In this way, if you are typing the producer, you can't use the sandbox mode unless you remove the typing. E.g.

```
class myClass() {
  public function __construct($producer, $message)
}
```

Adding the ProducerInterface you can type the producer and use both `Producer` and `Fallback`:

```
class myClass() {
  public function __construct(ProducerInterface $producer, $message)
}
```

### DIFF
--- a/RabbitMq/Fallback.php
+++ b/RabbitMq/Fallback.php
@@ -2,9 +2,9 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
-class Fallback
+class Fallback implements ProducerInterface
 {
-    public function publish()
+    public function publish($msgBody, $routingKey = '', $additionalProperties = array())
     {
         return false;
     }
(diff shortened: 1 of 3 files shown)


agiuliano:
This issue also solve #148

mbiagetti:
:+1:

videlalvaro:
Will this imply some BC breaking code?

agiuliano:
I don't think so, both classes have the `publish` method, you can either use the Producer as is or type hinting the variable against the interface.
In addition, tests seem ok

agiuliano:
@videlalvaro any chance of adding this in a tagged version?

videlalvaro:
Soon

PR #257, opened 2015-04-16
Allow consumer without exchange options

alex-pex:
I modified consumer configuration to allow consumers without exchange options

for example : 

```yaml
    producers:
        hello:
            connection:       default
    consumers:
        hello:
            connection:       default
            exchange_options: { name: 'amq.direct', type: direct }
            queue_options:    { name: 'hello' }
            callback:         hello_consumer
```

Can be writen like this : 
```yaml
    producers:
        hello:
            connection:       default
    consumers:
        hello:
            connection:       default
            queue_options:    { name: 'hello' }
            callback:         hello_consumer
```

In my case, the hello producer uses default exchange and direct queuing and the consumer exchange is useless

### DIFF
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -144,6 +144,9 @@ class OldSoundRabbitMqExtension extends Extension
             //this consumer doesn't define a queue, create a temporary queue
             if (!isset($consumer['queue_options'])) {
                 $consumer['queue_options']['name'] = '';
+                $consumer['queue_options']['durable'] = false;
+                $consumer['queue_options']['exclusive'] = true;
+                $consumer['queue_options']['auto_delete'] = true;
             }
             $definition->addMethodCall('setQueueOptions', array($this->normalizeArgumentKeys($consumer['queue_options'])));
             $definition->addMethodCall('setCallback', array(array(new Reference($consumer['callback']), 'execute')));
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -141,9 +141,9 @@ class OldSoundRabbitMqExtension extends Extension
                 $consumer['exchange_options']['declare'] = false;
             }
             $definition->addMethodCall('setExchangeOptions', array($this->normalizeArgumentKeys($consumer['exchange_options'])));
-            //this consumer doesn't define a queue
+            //this consumer doesn't define a queue, create a temporary queue
             if (!isset($consumer['queue_options'])) {
-                $consumer['queue_options']['name'] = null;
+                $consumer['queue_options']['name'] = '';
             }
             $definition->addMethodCall('setQueueOptions', array($this->normalizeArgumentKeys($consumer['queue_options'])));
             $definition->addMethodCall('setCallback', array(array(new Reference($consumer['callback']), 'execute')));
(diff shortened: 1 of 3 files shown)


alex-pex:
I updated my pull request to also allow consumer without queue options

This is an example use of my change : 

```yaml
    producers:
        hello:
            connection:       default
            exchange_options: { name: 'hello', type: fanout }
    consumers:
        hello:
            connection:       default
            exchange_options: { name: 'hello', type: fanout }
            callback:         hello_consumer
```

videlalvaro:
This library started as an implementation of certain messaging patterns. It has evolved over the years into a more generic tool, therefore you'll find things like the anon consumers

DependencyInjection/OldSoundRabbitMqExtension.php
```diff
@@ -131,12 +131,25 @@ protected function loadConsumers()
...
+            if (!isset($consumer['exchange_options'])) {
+                $consumer['exchange_options']['name'] = '';
+                $consumer['exchange_options']['type'] = 'direct';
+                $consumer['exchange_options']['passive'] = true;
+                $consumer['exchange_options']['declare'] = false;
+            }
+            $definition->addMethodCall('setExchangeOptions', array($this->normalizeArgumentKeys($consumer['exchange_options'])));
+            //this consumer doesn't define a queue, create a temporary queue
+            if (!isset($consumer['queue_options'])) {
+                $consumer['queue_options']['name'] = '';
```
videlalvaro: aren't these values the default ones here https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/BaseAmqp.php#L30 ?

videlalvaro:
Now the bundle has the variables `$exchangeDeclared` and `$queueDeclared`. I would introduce a new setter for them, so if exchange options or queue options are not present, they are set to true during initialisation from the `OldSoundRabbitMqExtension`. Then the code should avoid completely the exchange/queue declare and binding.

See: https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/BaseAmqp.php#L13

DependencyInjection/OldSoundRabbitMqExtension.php
alex-pex: No, they don't. I've made a diff between default values from BaseAmqp and values I needed for default transport. That's why I only override "passive" and "declare" options for example.

alex-pex:
I agree with you. I was overriding options because I saw something similar in the same file.

Maybe a better approach would be create 2 new classes "DefaultExchange" and "DefaultQueue" with dedicated options, like you did for https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/AnonConsumer.php#L14 ?

In this new class you could set declare to false like said previously

DependencyInjection/OldSoundRabbitMqExtension.php
videlalvaro: BTW, RabbitMQ doesn't allow declaring the default exchange.

videlalvaro:
Works for me.

alex-pex:
Instead of new classes I added two new methods : getDefaultQueueOptions and getDefaultExchangeOptions

When options are not defined, OldSoundRabbitMqExtension call these methods.

A new queue option was added "declare" with the same behavior as the "declare" option on the exchange.

One last thing : queue binding is not done anymore on the default exchange, as it is forbidden by RabbitMQ

See this new pull request : #273

PR #258, opened 2015-04-21
Default queueOptions[name] to null

naderman:
A producer doesn't need to specify any queue options. If you don't specify queue options the default value for `queueOptions[name]` is currently an empty string. Auto setup fabric then runs queueDeclare on the producer which checks whether `queueOptions[name] !== null` to see if it needs to set up a queue. This fails with empty string and a queue with a blank name - resulting in an automatically created name by rabbitmq - is created. This ends up creating tons of unwanted queues if you keep it running for a while.

So the fix is simple, default to null for the queue name.

### DIFF
--- a/RabbitMq/BaseAmqp.php
+++ b/RabbitMq/BaseAmqp.php
@@ -28,7 +28,7 @@ abstract class BaseAmqp
     );
 
     protected $queueOptions = array(
-        'name' => '',
+        'name' => null,
         'passive' => false,
         'durable' => true,
         'exclusive' => false,


videlalvaro:
By AMQP spec the default queue name is an empty string. 

If you want to avoid creating the queues once and again, then the place where the check is done should be improved/fixed, not the queue defaults

naderman:
Fair enough, does that make sense though? Why is there a null check at the moment? What would be the currently correct way not to declare a queue in a producer?

videlalvaro:
I can't remember why that's there :) probably to prevent doing a queue declare with the wrong options, but it's clear that PHP type juggling breaks the whole logic here.

So we have the `queueDeclared` protected member, we could hack around that to prevent a queue declare… maybe add a setter and rename it to `declareQueue` or something like that, and allow apps/config to set it to prevent declaring the queue all the time. 

BTW, why doesn't you app just set the queue name option?

naderman:
Well I simply don't want to define the queue in my producers? I have the definition in my consumers already, why would I have to duplicate the queue options in the producer config?

videlalvaro:
To answer your question: defining queues in producers/consumers really depend on the kind of messaging patter you are implementing. If you are sending images to resize later, your producers probably want to make sure the exchange, queue and bindings exist, to be sure messages won't be lost while routing. On the other hand, if you are sending notifications, then your producers probably don't need to care about declaring queues.

Back to the issue. I think the solution would be around checking if there are no queue arguments set on the config, then it should not declare the queue.

PR #261, opened 2015-05-05
-added --queue-name option to consumer commands

tiborb:
It works the same way as --route


videlalvaro:
What's the use case for this? Any docs?

tiborb:
It can be used for dynamic consumers. 
Say you have different users, each has it's own queue+consumer with the same callback and the number of users changes over time.
In this case, a consumer can be added for each (new) user on the fly without having to change the config.

```
rabbitmq:consumer consume_event --route=app.event.user_foo --queue-name=queue-user_foo
```

I could document the use case at the end on the "Consumers" section inside the README.

tiborb:
I will cancel this request because https://github.com/videlalvaro/RabbitMqBundle/issues/264 could solve this problem in a more elegant way.

PR #266, opened 2015-05-08
Added rabbitmq:dynamic-consumer command

tiborb:
A context-aware command: https://github.com/videlalvaro/RabbitMqBundle/issues/264

### DIFF
--- a/Command/BaseConsumerCommand.php
+++ b/Command/BaseConsumerCommand.php
@@ -87,14 +87,18 @@ abstract class BaseConsumerCommand extends BaseRabbitMqCommand
         if (0 > $this->amount) {
             throw new \InvalidArgumentException("The -m option should be null or greater than 0");
         }
-
+        $this->initConsumer($input);
+        
+        $this->consumer->consume($this->amount);   
+    }
+    
+    protected function initConsumer($input) {
         $this->consumer = $this->getContainer()
-            ->get(sprintf($this->getConsumerService(), $input->getArgument('name')));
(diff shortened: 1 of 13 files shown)


tiborb:
I can also imagine making the --context argument optional. 
In that case the command's usage would be like a "normal" consumer with the difference that the queue options are loaded only from a service which takes a context argument.

Command/BaseConsumerCommand.php
```diff
@@ -87,14 +87,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if (0 > $this->amount) {
             throw new \InvalidArgumentException("The -m option should be null or greater than 0");
         }
-
+        $this->initConsumer($input);
```
videlalvaro: why do you need these changes here?

tiborb: I have separated the consumer initialisation from the actual execution.

It is needed for extending the BaseConsumerCommand by the DynamicConsumerCommand and allows me to pass an argument to the dynamic consumer (the context in this case). 
It also makes the code more testable and easier to reuse.

videlalvaro: How does this interplay with other features. Does it introduce any BC breaks?

tiborb: No BC here.
Only the execute calls the init method, so there is no change in the standard consumer's  workflow.

videlalvaro:
thanks!

PR #269, opened 2015-05-20
Allow producer to not specify any queue options

alexgt9:
Related to the pull request https://github.com/videlalvaro/RabbitMqBundle/pull/258
This fixes the queue declaration without changing the default values of a queue. The queue should be declared only when configured.

### DIFF
--- a/RabbitMq/BaseAmqp.php
+++ b/RabbitMq/BaseAmqp.php
@@ -6,12 +6,18 @@ use PhpAmqpLib\Connection\AMQPConnection;
 use PhpAmqpLib\Connection\AMQPLazyConnection;
 
 abstract class BaseAmqp
-{
+{   
+    /**
+     * Queue name auto generated.
+     */
+    const QUEUE_NAME_AUTOGENERATE = '';
+
     protected $conn;
     protected $ch;
     protected $consumerTag;
     protected $exchangeDeclared = false;
     protected $queueDeclared = false;
+    protected $queueConfigured = false;
     protected $routingKey = '';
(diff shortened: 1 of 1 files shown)


RabbitMq/BaseAmqp.php
```diff
@@ -121,6 +127,7 @@ public function setExchangeOptions(array $options = array())
     public function setQueueOptions(array $options = array())
     {
         $this->queueOptions = array_merge($this->queueOptions, $options);
+        $this->queueConfigured = true;
```
videlalvaro: I think I prefer for this flag to be specifically set and not a side-effect of calling `setQueueOptions`

alexgt9: But doing in that way, we don't have the default behavior of declaring a exchange without queue

VasekPurchart:
@skafandri Why was this closed? I think the problem is stil valid (and really irritating). The solution proposed in #258 maybe wasn't the best but I think a suitable solution was proposed by @videlalvaro, isn't that right?

So if you don't like the PR then probably closing it is fine, but this issue is still valid.

VasekPurchart:
Or if it isn't can you please point me please towards what solved it?

skafandri:
https://github.com/php-amqplib/RabbitMqBundle/pull/273 doesn't fix it?

VasekPurchart:
Sorry for confusion, I found about about #273 only after I posted the messages (since the "PR-mention" here on GitHUb does not send a notification) I hope that does solve it :)

PR #272, opened 2015-05-21
Fixed consumers leak

remco1976:
basic_consume() returns a tag identifiying particular consumer, which we must supply to basic_cancel to remove this consumer. We also need to disable auto_delete attribute for the queue for it to remain available for subsequent use, after we create and then remove our consumer.

### DIFF
--- a/RabbitMq/RpcClient.php
+++ b/RabbitMq/RpcClient.php
@@ -44,13 +44,13 @@ class RpcClient extends BaseAmqp
     public function getReplies()
     {
         $this->replies = array();
-        $this->getChannel()->basic_consume($this->getQueueName(), '', false, true, false, false, array($this, 'processMessage'));
+        $consumer_tag = $this->getChannel()->basic_consume($this->getQueueName(), '', false, true, false, false, array($this, 'processMessage'));
 
         while (count($this->replies) < $this->requests) {
             $this->getChannel()->wait(null, false, $this->timeout);
         }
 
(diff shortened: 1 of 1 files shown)


RabbitMq/RpcClient.php
```diff
@@ -70,7 +70,7 @@ public function processMessage(AMQPMessage $msg)
     protected function getQueueName()
     {
         if (null === $this->queueName) {
-            list($this->queueName, ,) = $this->getChannel()->queue_declare("", false, false, true, true);
+            list($this->queueName, ,) = $this->getChannel()->queue_declare("", false, false, true, false);
```
videlalvaro: there are many ways to implement RPC one is with many clients per queue, which is what would happen if the queue is durable. In the bundles case it assumes one client that sends a request expects for the reply on it's own private queue, which means said queue is gone once the RPC Client is gone.

I would accept the other change about `consumer_tag` but not this one

remco1976: With auto_delete=true, queue will be gone right after the first rpc call with particular RpcClient instance (because count of counumers for the queue becomes 0 after we call basic_cancel with correct tag).
Turning off auto_delete won't make this queue live longer than client connection: RpcClient extends BaseAmqp, and BaseAmqp's __destruct() will close the connection if it is alive.
Because queue is exclusive, it will be deleted when connection closes. Therefore the queue cannot live longer than RpcClient instance.

remco1976:
and please note, that auto_delete=false is not the same as durable=true.

Issue #113, opened 2013-05-15
[Feature request] Support for configuring with an amqp uri


stof:
It would be great if the config could accept an [amqp uri](http://www.rabbitmq.com/uri-spec.html) and convert it internally to extract each setting. This would make it easier to use the bundle with hosted rabbitmq providing an amqp uri with the settings (for instance, CloudAMQP sets the uri directly as an environment variable on heroku)

t0k4rt:
:+1:

ramunasd:
Already implemented.

PR #271, opened 2015-05-26
Configuration of connection settings as URL

KingCrunch:
It would be nice, if I could configure the connection settings as a single string for example as an URL instead of 4 separate keys.

    //user:password@host:post/vhost

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,11 @@ old_sound_rabbit_mq:
 
             #requires php_sockets.dll
             use_socket: true # default false
+        another:
+            # A different (unused) connection defined by an URL. One can omit all parts,
+            # except the scheme (amqp:). If both segment in the URL and a key value (see above)
+            # are given the value from the URL takes precedence.
+            url: 'amqp://guest:password@localhost:5672/vhost?lazy=1&connection_timeout=6'
     producers:
         upload_picture:
             connection:       default
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -50,6 +50,7 @@ class Configuration implements ConfigurationInterface
                     ->canBeUnset()
                     ->prototype('array')
                         ->children()
+                            ->scalarNode('url')->defaultValue('')->end()
                             ->scalarNode('host')->defaultValue('localhost')->end()
                             ->scalarNode('port')->defaultValue(5672)->end()
                             ->scalarNode('user')->defaultValue('guest')->end()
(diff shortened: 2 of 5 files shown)


KingCrunch:
Hi @videlalvaro 

I've converted this one into a PR (I don't think I can do this with #113, can I?). Feedback appreciated. Especially one cannot set the `ssl_context`-parameters via URL. This solution is inspired by the way the doctrine-bundle handles the URL, so like there one can omit every part (here: except the scheme and the host) and in that case the normal configuration options will be used again.

README.md
```diff
@@ -120,6 +120,11 @@ old_sound_rabbit_mq:
 
             # requires php-amqplib v2.4.1+
             heartbeat: 0
+        another:
+            # A different (unused) connection defined by an URL. One can omit all parts,
+            # except the scheme (amqp:). If both segment in the URL and a key value (see above)
+            # are given the value from the URL takes precedence.
```
videlalvaro: why does the URL overrides what ones set above in a connection specifically? I would do it the other way around, the URL fills in values that might have not been specified in the connections confs

videlalvaro:
Are you following this spec: https://www.rabbitmq.com/uri-spec.html ?

KingCrunch:
Yes, at least I tried. Tell me, when I missed something. On the other hand the spec cover pretty common URIs, so there should be no bigger surprise.

videlalvaro:
Any replies to my comment:

>why does the URL overrides what ones set above in a connection specifically? I would do it the other way around, the URL fills in values that might have not been specified in the connections confs

videlalvaro:
@KingCrunch did you see my comments above? I would really like to merge your PR

KingCrunch:
@videlalvaro Sorry, I indeed missed that comment.

The reason the URL overwrites the other settings is, that this is the way the `DoctrineBundle` works. I don't know, what's the idea, but I wanted to keep it in sync. Both are bundles and having two connections URLs, but both behaves differently would be confusing. However, that's nothing I'd fight for.

videlalvaro:
@KingCrunch OK, if that's the way `DoctrineBundle` works, then we follow what the community is already doing. 

I'll review this again then.

kesar:
👍 it would be great to have this merged

dragosprotung:
Is there a reason this is not merged ?

KingCrunch:
At least there were some conflicts now. I've rebased the branch, so somebody can review again and merge afterwards.

PR #273, opened 2015-05-28
Added default exchange and queue options (v2)

alex-pex:
I rewrote the previous pull request #257 
It is also similar to #269 and #258

I modified consumer configuration to allow consumers without exchange options

for example : 

```yaml
    producers:
        hello:
            connection:       default
    consumers:
        hello:
            connection:       default
            exchange_options: { name: 'amq.direct', type: direct }
            queue_options:    { name: 'hello' }
            callback:         hello_consumer
```

Can be writen like this : 
```yaml
    producers:
        hello:
            connection:       default
    consumers:
        hello:
            connection:       default
            queue_options:    { name: 'hello' }
            callback:         hello_consumer
```

In my case, the hello producer uses default exchange and direct queuing and the consumer exchange is useless

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -277,6 +277,7 @@ class Configuration implements ConfigurationInterface
                 ->booleanNode('exclusive')->defaultFalse()->end()
                 ->booleanNode('auto_delete')->defaultFalse()->end()
                 ->booleanNode('nowait')->defaultFalse()->end()
+                ->booleanNode('declare')->defaultTrue()->end()
                 ->variableNode('arguments')->defaultNull()->end()
                 ->scalarNode('ticket')->defaultNull()->end()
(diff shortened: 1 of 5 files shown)


alex-pex:
I added two new methods : getDefaultQueueOptions and getDefaultExchangeOptions. When options are not defined, OldSoundRabbitMqExtension call these methods.

A new queue option was added "declare" with the same behavior as the "declare" option on the exchange.

One last thing : queue binding is not done anymore on the default exchange, as it is forbidden by RabbitMQ

zakjan:
What is the current status of this change?

videlalvaro:
@zakjan I just need to find time to review it

monsieurchico:
Actualy, it would be useful with amq.default as the exchange name (or no name)... amq.direct does not automaticaly binds the exchange and the queue

PR #276, opened 2015-06-20
Resolve default options for multiple consumer queues

agentsib:
Hi. Without this changes in config need set all values. Thanks.

### DIFF
--- a/RabbitMq/BaseConsumer.php
+++ b/RabbitMq/BaseConsumer.php
@@ -6,23 +6,43 @@ use PhpAmqpLib\Message\AMQPMessage;
 
 abstract class BaseConsumer extends BaseAmqp implements DequeuerInterface
 {
+    /** @var int */
     protected $target;
 
+    /** @var int */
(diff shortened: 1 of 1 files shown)


videlalvaro:
What's the idea? To set values just for some queues and then have others inherit values? 

Wouldn't this introduce unwanted consequences?

agentsib:
Hi. In base consumer you set default options for queue

https://github.com/videlalvaro/RabbitMqBundle/blob/5269d59e6d64a4a1b368d3890cbaaecdd82c12ca/RabbitMq/BaseAmqp.php#L121

But for multi consumer default options not set. I think it's logical.

For example. I use you bundle in silex provider.

https://github.com/agentsib/rabbitmq-service-provider/blob/master/src/fiunchinho/Silex/Provider/RabbitServiceProvider.php#L184

Without this changes I need check all queue params in config file. If you in future add some parameters it broken my code...

Thanks.

PR #280, opened 2015-06-24
RPC server option overwrites

Padam87:


Padam87:
Hi @videlalvaro , the current rpc server configuration determines the queue and the exchange names based on the key of the definition. This would be okay, but the current version of `setExchangeOptions` and `setQueueOptions` does not actually overwrites properties but merges them preferring existing ones. In short, you cannot overwrite these.

We need to configure the name of the parameters, and currently we cannot do that.

Pretty much any way I go about fixing this is a BC break.

What's your take on this one?

Padam87:
My preferred solution is to remove the array_merge here: https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/BaseAmqp.php#L114 and here: https://github.com/videlalvaro/RabbitMqBundle/blob/master/RabbitMq/BaseAmqp.php#L123

This could only be included in a new major release unfortunately.

Issue #121, opened 2013-06-20
one producer, one exchange, multiple queues


ryanalbon:
Hi there,

I'm having trouble understanding how to bind multiple queues to one exchange via a single producer. I understand that I can bind a single queue to an exchange with a configuration that looks like this:

```
producers:
    producer_1:
        connection: default
        exchange_options: {name: 'exchange_1', type: direct}
        queue_options: {name: 'queue_1', routing_keys: ['key_1']}
```

The above will configure a producer called "producer_1" which publishes to "exchange_1", and binds "exchange_1" to "queue_1" via "key_1".

I'd like to use the same producer to bind multiple queues to one exchange via different routing keys. I was hoping something like the following would work:

```
producers:
    producer_1:
        connection: default
        exchange_options: {name: 'exchange_1', type: direct}
        queue_options:
            - {name: 'queue_1', routing_keys: ['key_1']}
            - {name: 'queue_2', routing_keys: ['key_2']}
            - {name: 'queue_n', routing_keys: ['key_n']}
```

But as far as I can tell, the bundle doesn't support any kind of configuration that binds multiple queues to one exchange under one producer.

So, my question is: is it possible to configure one producer which binds multiple queues to one exchange? Or do I have to configure one producer per queue?

Thanks very much for your help!

Ryan

iLukey:
@bastnic that's true, but in that case I don't think that the config should allow *any* queue options for a producer at all. Producers should be blissfully unaware of anything beyond the exchange that their message is going to. Perhaps the bundle could use the logger to generate a warning if there are no queues configured for the exchange, but those queues should've been defined as part of the consumer config anyway imo.

fatmuemoo:
I agree that this scenario is missing from the current configuration. I have a use case for one queue, multiple exchanges.  Would the maintainers consider a PR that introduces multiple bindings?

videlalvaro:
I would consider a PR but keep in mind this bundle started as a way to show/implement messaging patterns. For some cases you want the producer to define a queue, in other cases you don't need that.

Example: **image resize**.  If your producer doesn't define the queue and the binding, you effectively drop users request in the case the consumer is not already running, so you want to setup that in advance.

Example: **logging**. You are sending logs/events via RabbitMQ, then you don't need the producer to set up the queue/bindings. Anyone interested in those will start a consumer and perform the required bindings.

So, yes, things can be improved, but in messaging not everything is black/white.

Issue #36, opened 2012-03-23
Multiple Bindings


aanchalgera:
Using this bundle, How can I bind multiple routing keys to a queue?

http://www.rabbitmq.com/tutorials/tutorial-four-python.html - Multiple bindings

Something like this:-
for severity in severities:
    channel.queue_bind(exchange='direct_logs',
                       queue=queue_name,
                       routing_key=severity)

Issue #30, opened 2012-03-01
Consumer architecture


pluff:
Hello @videlalvaro 

I review your bundle a bit and I have couple of issues.
1. Does producer should know anything about queues? As far as I understand producer just publishes messages to exchange. The only thing producer should declare is an exchange.
2. Does consumer should know anything about exchanges and routing keys? As far as I understand consumer just consumes the message from a queue and doesn't care where it come from. The only thing consumer should declare is a queue he consumes.
3. If both questions above should be answered "no" then we need a way to setup exchanges,queues and bindings between them and here comes my issue #27. I can implement these changes, but I just need your opinion maybe I have missed something important.

castarco:
Why there are open issues since 2012 and with no response?

videlalvaro:
@castarco replied in another issue. Cheers

PR #281, opened 2015-06-26
Arbitrary Bindings

fatmuemoo:
You may find that your application has a complex workflow and you you need to have arbitrary binding. Arbitrary binding scenarios might include exchange to exchange bindings via `destination_is_exchange` property.

```yaml
bindings:
    - {exchange: foo, destination: bar, routing_key: 'baz.*' }
    - {exchange: foo1, destination: foo, routing_key: 'baz.*' destination_is_exchange: true}
```

The rabbitmq:setup-fabric command will declare exchanges and queues as defined in your producer, consumer and multi consumer configurations before it creates your arbitrary bindings. However, the rabbitmq:setup-fabric will *NOT* declare addition queues and exchanges defined in the bindings. It's up to you to make sure exchanges/queues are declared.

### DIFF
--- a/Command/SetupFabricCommand.php
+++ b/Command/SetupFabricCommand.php
@@ -28,8 +28,11 @@ class SetupFabricCommand extends BaseRabbitMqCommand
 
         $partsHolder = $this->getContainer()->get('old_sound_rabbit_mq.parts_holder');
 
-        foreach ($partsHolder->getParts('old_sound_rabbit_mq.base_amqp') as $baseAmqp) {
-            $baseAmqp->setupFabric();
+        foreach (array('base_amqp', 'binding') as $key) {
+            foreach ($partsHolder->getParts('old_sound_rabbit_mq.' . $key) as $baseAmqp) {
+                $baseAmqp->setupFabric();
+            }
         }
+
     }
 }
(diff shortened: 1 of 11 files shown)


fatmuemoo:
This will fix #121 and possibly #30 and #36

fatmuemoo:
CI errors are due to composer failing to download itself

videlalvaro:
This new bindings keys is just used for SetupFabric, right?

Also, as far as I understand the code, the `bindings` config key is required, right? Can it be made optional for BC?

fatmuemoo:
Yes, these are only used for setup fabric. It was my intention not to break backwards compatible, I'll double check and make the bindings config key optional.

fatmuemoo:
If I understand the [Symfony Documentation](http://symfony.com/doc/current/components/config/definition.html#default-and-required-values), it is optionsal because I never call isRequried or cannotBeEmpty. Confirmed with 2.5.9 and 2.7 that it is only required if I add isRequired in Configuration::addBindings

ggnet:
+1 for this feature. We use exchange to exchange binding

ggnet:
@videlalvaro Hi. Do you plan to merge this PR? We really need this feature.
Thank you!

ChristianRiesen:
Ping @videlalvaro 
We could make use of this feature as well.

kasparsklavins:
Ping @videlalvaro

videlalvaro:
@kasparsklavins there should be new maintainers for the bundle

PR #282, opened 2015-06-30
Fixed producer fallback in sandbox mode

ramunasd:
Fallback lacks method setContentType

### DIFF
--- a/RabbitMq/Fallback.php
+++ b/RabbitMq/Fallback.php
@@ -2,10 +2,24 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+/**
+ * Fallback producer for sandbox mode.
+ */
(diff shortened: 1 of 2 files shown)


videlalvaro:
The tests are not passing.

Also, wouldn't this break BC?

ramunasd:
Travis had temporary issue, now it's passing all tests.
About BC - if someone has written it's own producer that implements ProducerInterface - then yes, it BC.

videlalvaro:
Sorry to comment so late, but I think this method and `setDeliveryMode` should be added here as well? Why not have them in the ProducerInterface as well?

ramunasd:
@mihaileu this is the old one, but maybe still actual?

mihaileu:
As is a BC, it should be added in the next major version.

PR #295, opened 2015-08-19
Removed reliance on deprecated class

martijn4evers:
Since Aug 5, 2015 AMQPLazyConnection does not extend AMQPConnection anymore in php-amqplib (https://github.com/videlalvaro/php-amqplib/commit/26ef3810a68ee60a976b0ccfb64a2bd574e66908). AbstractConnection exist since 14 Feb 2012, so I do not expect any BC problems.


videlalvaro:
thanks

PR #299, opened 2015-09-05
Correct exchange name in RPC example.

piepkrak:
The first RPC example does not work without changing the exchange name of the RPC server.


videlalvaro:
thanks

piepkrak:
You're welcome!

PR #303, opened 2015-09-28
Support PhpAmqpLib 2.6

ruudk:


videlalvaro:
@ruudk there are failing tests

ruudk:
I see, but I don't know how to fix that one. 

Does it actually connect to socket localhost:123? I don't have that on my system. How did the previous test handle that?

ruudk:
Tests are fixed :)

Tests/RabbitMq/BaseAmqpTest.php
```diff
@@ -9,7 +9,7 @@
 class BaseAmqpTest extends \PHPUnit_Framework_TestCase
 {
     /**
-     * @expectedException \PhpAmqpLib\Exception\AMQPRuntimeException
+     * @expectedException \ErrorException
```
videlalvaro: why do you have to change the exception type here?

composer.json
```diff
@@ -14,7 +14,7 @@
         "symfony/config"               : "~2.3",
         "symfony/yaml"                 : "~2.3",
         "symfony/console"              : "~2.3",
-        "videlalvaro/php-amqplib"      : "~2.5.0"
+        "videlalvaro/php-amqplib"      : "~2.6"
```
videlalvaro: I think this is the only relevant change of this PR

Tests/RabbitMq/BaseAmqpTest.php
ruudk: Doesn't work. The exception is thrown here: https://github.com/videlalvaro/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L301

videlalvaro: I think that might be because another kind of error is being thrown so it ends up there. Anyway, I've merged previous PRs that address this already. I think the only thing to do would be to update composer.json

ruudk: Unfortunately no. I ran updated the composer.json on master and ran phpunit and it gave the same error. So I don't know which exception we are trying to catch here, because the test doesn't say that. What to do?

ruudk: Ok, I found out that the original test was trying to catch this exception in v2.4.1 https://github.com/videlalvaro/php-amqplib/blob/58044157b30853a1a425e26e9ff9895247668ac3/PhpAmqpLib/Wire/IO/StreamIO.php#L104

Now, the error originates from here:
https://github.com/videlalvaro/php-amqplib/blob/8a6e89ad46130eb365b7f57d313f2a795f5e3269/PhpAmqpLib/Wire/IO/StreamIO.php#L115
and here
https://github.com/videlalvaro/php-amqplib/blob/8a6e89ad46130eb365b7f57d313f2a795f5e3269/PhpAmqpLib/Wire/IO/StreamIO.php#L301

What to do?

ruudk: In the old version stream_socket_client was suppressing errors with `@`. So my test is the way to catch the exception :)

videlalvaro:
merged, thanks

PR #307, opened 2015-10-25
Updated composer instructions to match best practice

ricbra:
It's recommended to install packages with <code>composer require</code> instead of manually editing the <code>composer.json</code> file.

The package will be bind to the latest version with <code>^</code> which is also recommended.


videlalvaro:
thanks

PR #345, opened 2016-03-18
Allow setting headers on publishing message

dakorpar:
This will enable setting headers when publishing message.
For example it's going to be possible to send delayed message with setting x-delay header (you need x-delayed-message plugin for RabbitMq)

### DIFF
--- a/RabbitMq/Producer.php
+++ b/RabbitMq/Producer.php
@@ -4,6 +4,7 @@ namespace OldSound\RabbitMqBundle\RabbitMq;
 
 use OldSound\RabbitMqBundle\RabbitMq\BaseAmqp;
 use PhpAmqpLib\Message\AMQPMessage;
+use PhpAmqpLib\Wire\AMQPTable;
 
 /**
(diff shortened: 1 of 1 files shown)


RabbitMq/Producer.php
```diff
@@ -38,14 +39,19 @@ protected function getBasicProperties()
...
-    public function publish($msgBody, $routingKey = '', $additionalProperties = array())
+    public function publish($msgBody, $routingKey = '', $additionalProperties = array(), array $headers = null)
     {
         if ($this->autoSetupFabric) {
             $this->setupFabric();
         }
 
         $msg = new AMQPMessage((string) $msgBody, array_merge($this->getBasicProperties(), $additionalProperties));
+
+        $headersTable = new AMQPTable($headers);
```
stloyd: You should set headers only if it were passed to that method.

dakorpar: AMQPAbstractCollection constructor which AMQPTable inherits already has this check so I skipped it, but I can add check here as well.

PR #346, opened 2016-03-29
Make :purge and :delete commands compatible with Symfony 3.0

PapyDanone:
Replaces `Dialog` helper calls with `ConfirmationQuestion`. Without it, they fail if you are not using the `--no-confirmation` flag. See deprecation notice in [symfony docs](http://symfony.com/doc/current/components/console/helpers/dialoghelper.html).


PapyDanone:
The only problem is that the `ConfirmationQuestion` class was introduced in `symfony/console ~2.7`. Should we therefore bump the version in composer?

PR #348, opened 2016-04-07
Force fetching new channel when channel_id is null

### LINKED ISSUE #347: Dead channel when performing a reconnect.
In a producer process, calling reconnect() function breaks the connection of current channel object.

Here is my code :
```php
    $this->producer->reconnect();
    $this->producer->publish(json_encode($m));
```

And what I obtain :
```
Fatal error: Call to a member function send_channel_method_frame() on a non-object in /home/users/terrien/my-project/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php on line 223
```

I fix this by edditing the getChannel function :
```php
    public function getChannel()
    {
        if (empty($this->ch) || null === $this->ch->getChannelId()) {
            $this->ch = $this->conn->channel();
        }

        return $this->ch;
    }
```

What do you think ?

sterrien:
Fix #347

### DIFF
--- a/RabbitMq/BaseAmqp.php
+++ b/RabbitMq/BaseAmqp.php
@@ -80,7 +80,7 @@ abstract class BaseAmqp
      */
     public function getChannel()
     {
-        if (empty($this->ch)) {
+        if (empty($this->ch) || null === $this->ch->getChannelId()) {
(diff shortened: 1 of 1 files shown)


sterrien:
Thanks a lot !

PR #362, opened 2016-06-02
Make processMessage method abstract

atailouloute:
Hello,
I noticed that the `processMessage` in the abstract class `BaseConsumer` does not have any implementation, it will be good to make the method abstract as the class itself is abstract.
Regards,


skafandri:
For the moment, this would introduce a BC break since many people may had extended `BaseConsumer` and implemented `processMessage` with a different signature, example:
````
    public function processMessage($msg)
    {
        //custom implementation
    }
````
Therefore, this will make it into 2.0 version.
I will create a 2.0 branch and an issue with a change-set draft, so everybody can join the discussion.
I will ping here when done.

atailouloute:
I forgot this point, but it will be nice to add it as you said in a new branch.

:+1:

PR #365, opened 2016-06-08
Add shutdown method. When called kernel shutdown, we must close opene…

shustrik:
When kernel shutdown is called, we must close opened connections. For example behat with symfony2 extension uses kernel shutdown for each scenario.

### DIFF
--- a/OldSoundRabbitMqBundle.php
+++ b/OldSoundRabbitMqBundle.php
@@ -16,4 +16,17 @@ class OldSoundRabbitMqBundle extends Bundle
         $container->addCompilerPass(new RegisterPartsPass());
         $container->addCompilerPass(new InjectEventDispatcherPass());
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function shutdown()
+    {
+        parent::shutdown();
+        $partHolder = $this->container->get('old_sound_rabbit_mq.parts_holder');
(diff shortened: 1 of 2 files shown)


OldSoundRabbitMqBundle.php
```diff
@@ -14,4 +14,17 @@ public function build(ContainerBuilder $container)
...
+    /**
+     * {@inheritDoc}
+     */
+    public function shutdown()
+    {
+        parent::shutdown();
+        $partHolder = $this->container->get('old_sound_rabbit_mq.parts_holder');
+        $connections = $partHolder->getParts("old_sound_rabbit_mq.base_amqp");
+        foreach ($connections as $connection) {
+            $connection->__destruct();
```
shustrik: or extract method closeConnection() in BaseAmqp

RabbitMq/BaseAmqp.php
```diff
@@ -67,23 +67,20 @@ public function __construct(AbstractConnection $conn, AMQPChannel $ch = null, $c
...
+
+    public function close()
+    {
         if ($this->ch) {
-            try {
-                $this->ch->close();
-            } catch (\Exception $e) {
-                // ignore on shutdown
-            }
+            $this->ch->close();
```
skafandri: Why you removed the try/catch ?

shustrik: My mistake. Add try catch

PR #368, opened 2016-06-23
Events

alexandrupanturu:
It would be nice to have some events on consume and before/after processing a message.
Example:
You can listen to "onConsume" event and can sleep consumer on new app deploy


DependencyInjection/Compiler/InjectEventDispatcherPass.php
```diff
@@ -0,0 +1,37 @@
...
+            foreach ($taggedConsumers as $id => $tag) {
+                $definition = $container->getDefinition($id);
+                $definition->addMethodCall(
+                    'setEventDispatcher',
+                    [$eventDispatcherDefinition]
+                );
+            }
+
+        } catch (\Exception $e) {
+            // DO Nothing
```
skafandri: Would be better to check if an event_dispatcher is available rather than catching \Exception. We may hide other exceptions others than no event_dispatcher within the container.

Event/Consumer/AbstractConsumerEvent.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+abstract class AbstractConsumerEvent extends Event
+{
+    /**
+     * Event Name
+     *
+     * @return string
+     */
+    abstract public function getName();
+}
```
skafandri: `Symfony\Component\EventDispatcher\Event::getName` is deprecated since version 2.4, to be removed in 3.0
Maybe have a single `ConsumerEvent` with the event names as constants?

alexandrupanturu: Will fix that. Wrote it on a symfony 3.0 and did not catch the deprecation.

PR #374, opened 2016-07-18
Heartbeat implementation on RabbitMQ consumers.

tiagoblackcode:
In a project I'm working on we have several problems regarding the persistent trait of RabbitMQ consumers. Specifically, when the queue a consumer is listening on has no messages for a while and the consumer is dependant on a MySQL connection, the next message it receives will raise an exception with the message "MySQL has gone away" since MySQL closes any idle connection that sits idle for longer than `wait_timeout`.

To prevent any dependencies between the RabbitMQ bundle and MySQL I thought about having a `heartbeat` function that consumers _can_ implement to perform some periodic tasks (such as refreshing the MySQL socket).

This enables the use of a `heartbeat_timeout` parameter in the consumer configuration that will periodically call the `heartbeat($consumerDuration)` function on the consumer if it implements the `HeartbeatAwareConsumerInterface`. This is being done by passing the `heartbeatTimeout` value into the `$channel->wait(...)` method and checking if we are in a `idleTimeout` or `heartbeatTimeout` situation. If it's the former the exception thrown by the channel is re-raised, otherwise it be swallowed and the consumer will be notified.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -137,6 +137,7 @@ class Configuration implements ConfigurationInterface
                             ->scalarNode('connection')->defaultValue('default')->end()
                             ->scalarNode('callback')->isRequired()->end()
                             ->scalarNode('idle_timeout')->end()
+                            ->scalarNode('heartbeat_timeout')->end()
                             ->scalarNode('auto_setup_fabric')->defaultTrue()->end()
                             ->arrayNode('qos_options')
                                 ->canBeUnset()
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -193,6 +193,11 @@ class OldSoundRabbitMqExtension extends Extension
                 ));
             }
 
+            if (isset($consumer['heartbeat_timeout']) && $consumer['heartbeat_timeout'] > 0) {
+                $definition->addMethodCall('setHeartbeatTimeout', array($consumer['heartbeat_timeout']));
+                $definition->addMethodCall('setHeartbeatCallback', array(array(new Reference($consumer['callback']), 'heartbeat')));
+            }
+
             if (isset($consumer['idle_timeout'])) {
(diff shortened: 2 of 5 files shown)


skafandri:
I think the main goal here is to be able to *heartbeat* some external services like a database server.
Since we already have a **heartbeat** configuration to heartbeat the RabbitMQ server I suggest to use the same moment and heartbeat whatever we want.
I think the proposed solution is a little bit over complicated for users. I will have to update all my consumers configuration, implement an extra interface..

What about dispatching an event after heartbeating RabbitMq server, so anyone can just register an event listener and do whatever they want?

tiagoblackcode:
Hello @skafandri,

thanks for your feedback!

I looked over at how using the socket hearbeat feature would work and, despite the approach being desirable, we would have to re-implement the `PhpAmqpLib\Connection\AbstractConnection` and all its sub-classes on this bundle to trigger the event (the targeted method is `wait_channel($channel_id, $timeout = 0)`). This creates a dependency on the connection class implementation provided by `PhpAmqpLib`.

The current implementation only requires an additional key on the consumers configuration, i.e.

```yaml
consumers:
        my_consumer:
            ...
            heartbeat_timeout: 60
```

Instead of using the `HeartbeatAwareConsumerInterface` interface we can use the event dispatcher and keep the consumer heartbeat feature (something like `on_consumer_heartbeat`). This will make it easier for users that want to opt-in to the consumer heartbeat requiring only the addition of the `heartbeat_timeout` key to the consumers and listening to the `on_consumer_heartbeat` event.

Something I forgot to say is that the proposed implementation is backwards-compatible.

ramunasd:
I went through implementation and it seems too complicated. I would suggest to do another, not heartbeats focused approch:
let's name it "idle callbacks" and call them whenever timeout exception occurs, without checking time and etc. It's up to users what to implement in those "idle callbacks" because there might be very different heartbeat checks or so.

ramunasd:
My bad, idle callbacks are already there. No need to change this place.

tiagoblackcode:
Hello,

thank you for your feedback @ramunasd. 
It's been a while and I am no longer working on the project where this was needed but AFAIK it's still using the fork.

PR #375, opened 2016-07-21
Logger was not working properly

alexbumbacea:
Logger was not working properly.


skafandri:
What about adding the check [here](https://github.com/alexbumbacea/RabbitMqBundle/blob/6f1056baa6634d91d5416001b4f0c4c116c707f0/DependencyInjection/OldSoundRabbitMqExtension.php#L47)
So it will be checked only once

alexbumbacea:
The problem is that inside the "Extension" class the service "logger" is not **yet** available. Checking it there would be the same issue. The proposed fix, will simply not call the setter if the "@logger" is not available.

alexbumbacea:
@skafandri can we merge this?

Checking for the logger at that point can not be done as the definition of logger is not yet loaded into the container. The proposed solution is the optimal i have found.

Issue #377, opened 2016-08-02
Performance issue when closing all connection on shutdown of the bundle.


alexbumbacea:
https://github.com/php-amqplib/RabbitMqBundle/commit/d58390da78d3db49709ab4b82800ccd42121f044

This is not ok. Given a lazy connection, if you don't use in current request rabbit at all.... $connection will be initialized here as it is a lazy service not required until now...

PR #378, opened 2016-08-02
issue #377

shustrik:
issue #377.
Add parameter contain base_amqp connections. Because when get old_sound_rabbit_mq.parts_holder initialized all connections which added to part holder


alexbumbacea:
Guys? When will this be merged to master?

PR #395, opened 2016-10-18
Added support for connection URL encoding

dragosprotung:
With the introduction of the URL connection encoding was not taken into consideration.
See https://www.rabbitmq.com/uri-spec.html for more details.

Without this is not possible to connect with a vhost that does not start with '/'


stloyd:
Merged, thanks!

PR #398, opened 2016-10-19
Dynamic Connection Parameters

davidcochrum:
Sometimes your connection information may need to be dynamic. Dynamic connection parameters allow you to supply or override parameters programmatically through a service.

e.g. In a scenario when the `vhost` parameter of the connection depends on the current tenant of your white-labeled application and you do not want (or can't) change it's configuration every time.

Define a service under `connection_parameters_provider` that implements the `ConnectionParametersProviderInterface`, and add it to the appropriate `connections` configuration.

``` yaml
connections:
    default:
        host:     'localhost'
        port:     5672
        user:     'guest'
        password: 'guest'
        vhost:    'foo' # to be dynamically overridden by `connection_parameters_provider`
        connection_parameters_provider: connection_parameters_provider_service
```

Example Implementation:

``` php
class ConnectionParametersProviderService implements ConnectionParametersProvider {
    ...
    public function getConnectionParameters() {
        return array('vhost' => $this->getVhost());
    }
    ...
}
```

In this case, the `vhost` parameter will be overridden by the output of `getVhost()`.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -67,6 +67,7 @@ class Configuration implements ConfigurationInterface
                             ->end()
                             ->booleanNode('keepalive')->defaultFalse()->info('requires php-amqplib v2.4.1+ and PHP5.4+')->end()
                             ->scalarNode('heartbeat')->defaultValue(0)->info('requires php-amqplib v2.4.1+')->end()
+                            ->scalarNode('connection_parameters_provider')->end()
                         ->end()
                     ->end()
                 ->end()
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -4,7 +4,6 @@ namespace OldSound\RabbitMqBundle\DependencyInjection;
 
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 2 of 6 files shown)


mvrhov:
IMO, there is no need anymore, the symfony 3.2 supports environment variables which are resolved at runtime. So All you would need to set is an environment variable in nginx/apache

davidcochrum:
In my case, I needed the vhost to be dynamically provided by a service. That is not possible via environment variable.

crussell52:
@mvrhov  Thank you for your feedback.  Unfortunately, our org is not yet running symfony 3.2, but we'll be looking into the functionality you describe.

However, I do not believe that it obviates this feature.  Using an environment variable assumes that the connection configuration is embedded in the URL. In white-label or multi-tenant applications, it is more likely that connection settings need to be loaded from the database based on who is authenticated or what site is being accessed.  Any sort of lookup would be outside the scope of the web-server's responsibility. 

As a more concrete example, consider dynamic credentials. It seems reasonable to want to use roll separate credentials for each vhost to mitigate the risk of a successful user attempt to coerce the active vhost.  Trying to deriving credentials via apache/nginx would be an inappropriate solution.

crussell52:
@mvrhov 

Thank you for your earlier suggestion of using environment variables with Symfony 3.2... Unfortunately, that did not solve the problem. 😞 

The short version is that the configuration is stored in a database so it requires code to extract it.  This PR meets that need.

Are there other concerns about this feature?  
Are there another way to do this within this library that we've overlooked?

lombartec:
Just was looking for this feature in the docs, would be really nice to have.

crussell52:
Any progress on this?

This enhancement implementation appears to:
  * Be backwards compatible
  * Have test coverage
  * Solve a valid use case (dynamic connection parameters -- e.g. those stored in a database)
  * Be consistent with existing features (`QueuesProviderInterface`, `QueueOptionsProviderInterface`) 

Is there anything additional we can contribute to help this merge go through?

PR #402, opened 2016-11-03
Do not create channel for lazy socket connection on construct

munkie:
Now when using lazy socket connection and class extending BaseAmqp is initiated, channel is created.
Thou when lazy stream connection is used, channel would not be initialized.
This PR fixes this bug.


stloyd:
Looks good, thanks!

Issue #372, opened 2016-07-14
\OldSound\RabbitMqBundle\Command\BaseConsumerCommand::stopConsumer has exit() call


alexbumbacea:
Hello,

We are encountering a issue in usage of this consumer. In the stopConsumer method there is an exit() call.

This was noticed while sending SIGTERM to the process while a message was processed and the "console.terminate" listener was not executed. It is not proper to call exit inside a library.

The commit that introduced this was this: https://github.com/php-amqplib/RabbitMqBundle/commit/4f6a5c86c7dabf5fee3aa7f9e656998ef90fa8cd

Should we patch this inside the core library or should we overwrite in our project?

alexbumbacea:
in merge request we should also consider #352

Issue #260, opened 2015-05-05
[Consumer] Doesn't respond to SIGINT when there are no new messages


ruudk:
So I noticed that the Consumer can only be stopped when a new message arrives (#89). I think this is because of this line https://github.com/videlalvaro/RabbitMqBundle/blob/b546af972552a43c1fd3e23226df793b6f709d5e/RabbitMq/Consumer.php#L47

So I was thinking about how to fix this. What if we'd run that `while` with a AMQPTimeout inside another `while(true)`? This is how PHP Resque solves this.

Something like this:

``` php
    public function consume($msgAmount)
    {
        $this->target = $msgAmount;

        $this->setupConsumer();

        while(true) {
            $this->maybeStopConsumer();

            try {
                while (count($this->getChannel()->callbacks)) {
                    $this->maybeStopConsumer();
                    $this->getChannel()->wait(null, false, $this->getIdleTimeout());
                }
            } catch(\PhpAmqpLib\Exception\AMQPTimeoutException $exception) {
                echo "Timeout, restart\n";
            }
        }
    }
``
```

videlalvaro:
Doesn't this handle your problem? https://github.com/videlalvaro/RabbitMqBundle/commit/3e211851d25a9670ed0a3e175a0ba217e4708ab0

ruudk:
That works :) Thanks
Didn't see that fixed in master, sorry!

But whenever I kill the consumer with Control+C it throws an exception:
```
                                                              
  [PhpAmqpLib\Exception\AMQPRuntimeException]                 
  Error Connecting to server (36): Operation now in progress  
                                                              
Exception trace:
 () at /www/pdf-services/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Wire/IO/StreamIO.php:106
 PhpAmqpLib\Wire\IO\StreamIO->connect() at /www/pdf-services/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php:195
 PhpAmqpLib\Connection\AbstractConnection->connect() at /www/pdf-services/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php:179
 PhpAmqpLib\Connection\AbstractConnection->__construct() at /www/pdf-services/vendor/videlalvaro/php-amqplib/PhpAmqpLib/Connection/AMQPStreamConnection.php:41
[...]

PR #412, opened 2016-12-06
Fix for consumer with empty queue not being able to handle SIGTERM and SIGINT

EllisV:
Whenever sending a SIGTERM or SIGINT to a consumer it will only be able to shut itself down on a next message. But this might be problematic for queues which are not that active. Previously this has been solved by having an exit statement after `BaseConsumerCommand::stopConsumer()` method (ref: https://github.com/php-amqplib/RabbitMqBundle/commit/4f6a5c86c7dabf5fee3aa7f9e656998ef90fa8cd).

This implementation is done referencing to https://github.com/php-amqplib/php-amqplib/pull/224 and this example [amqp_consumer_signals.php#L179](https://github.com/roihq/php-amqplib/blob/ddb5b0374f34a454cd8f772f9a12f147e8c33e30/demo/amqp_consumer_signals.php#L179)

The only thing that I am concerned is that I think we are no longer waiting for `basic.cancel_ok` signal. Are there any possible scenarios that this could go wrong?

This solution still dispatches "console.terminate" as requested in #372. Doesn't do an exit call like it was previously implemented.

Fix for #352 #178 #260

### DIFF
--- a/RabbitMq/BaseConsumer.php
+++ b/RabbitMq/BaseConsumer.php
@@ -34,9 +34,15 @@ abstract class BaseConsumer extends BaseAmqp implements DequeuerInterface
         }
     }
 
+    /**
+     * Tell the server you are going to stop consuming.
+     *
+     * It will finish up the last message and not send you any more.
+     */
     public function stopConsuming()
     {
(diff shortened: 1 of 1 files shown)


stloyd:
Looks good, thanks!

Issue #403, opened 2016-11-07
[Feature] Bulk Consumer


tiborb:
I suggest the implementation of a bulk consumer. 
It would load a batch of message from RabbitMQ, and ack them all at once when the task is complete.
The batch size and timeout would be configurable.

Example usage for bulk loading data into a database.
Further reading: http://www.rabbitmq.com/consumer-prefetch.html

Any thoughts?

-Tibor

skafandri:
using the prefetch feature, prefetch_count messages will be loaded in your process memory. Then acked/nacked one by one. So the "batch" loading behavior is already implemented.
Is the batch acking/nacking worth the effort?

tiborb:
Inserting multiple values (bulk) into a database, e.g column store, and ack onfy if the db insert was sucessful.

devrck:
I would like to give it a shoot, because funny enough we implemented something similar :laughing:

tiborb:
Awesome, would like to hear about your implementation :+1:

devrck:
When you guys have the time please review my initial PR.

PR #413, opened 2016-12-07
[Feature] Bulk Consumer

devrck:
#403 implementation

* This new type of consumer gets x messages from the queue in unacknowledged state 
* Executes for each of them the **@execute()** method
* When the batch is complete or it's timeout wait is fulfilled then it will fire the method **@batchExecute()**
* If the **@batchExecute()** is without any exceptions it will acknowledged the messages if not it will re-queue them to be handled by another worker

[ **Obs** ]: Message can get stuck on the queue forever because of the behavior of the **@batchExecute()** when it has exceptions. I am open for suggestions on how to fix this.

Configuration used in app/config/config.yml
```yaml
# RabbitMQ Configuration
old_sound_rabbit_mq:
    connections:
        default:
            host:     '%rabbit_mq_conn_default_host%'
            port:     '%rabbit_mq_conn_default_port%'
            user:     '%rabbit_mq_conn_default_user%'
            password: '%rabbit_mq_conn_default_password%'
            vhost:    '%rabbit_mq_conn_default_vhost%'
            lazy:     '%rabbit_mq_conn_default_lazy%'
    batch_consumers:
        devck_basic:
            connection:       default
            exchange_options: {name: '%rabbit_mq_exchange_devck%', type: fanout}
            queue_options:    {name: '%rabbit_mq_queue_devck%'}
            callback:         devck.basic
            qos_options:      {prefetch_size: 0, prefetch_count: 2, global: false}
            timeout_wait:     5
            auto_setup_fabric: false
            idle_timeout_exit_code: -2
```

Callback class

```php
<?php

namespace AppBundle\Service;

use OldSound\RabbitMqBundle\RabbitMq\BatchConsumerInterface;
use PhpAmqpLib\Message\AMQPMessage;

class DevckBasicConsumer implements BatchConsumerInterface
{
    public function batchExecute()
    {
        echo 'Doing batch execution', PHP_EOL;
    }

    /**
     * @inheritDoc
     */
    public function execute(AMQPMessage $msg)
    {
        echo $msg->body, PHP_EOL;

        return true;
    }
}
```


skafandri:
How I access the messages in `batchExecute()` ?

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @return  void
+     *
+     * @throws  AMQPTimeoutException
+     */
+    private function consumeMessage($timeout)
+    {
+        $this->getChannel()->wait(null, false, $timeout);
+    }
+}
```
skafandri: new line :)

Command/BatchConsumerCommand.php
```diff
@@ -0,0 +1,20 @@
...
+            ->setName('rabbitmq:batch-consumer')
+            ->setDescription('Executes a batch consumer')
+        ;
+    }
+
+    protected function getConsumerService()
+    {
+        return 'old_sound_rabbit_mq.%s_batch';
+    }
+}
```
skafandri: new line :)

devrck:
Well my premise is that you do work in the **execute()** part and then in the **batchExecute()** you will do a action that is very costly.

Take by example a batch process for inserting in a database using Doctrine. You do the work in **execute()**, persisting objects, modifying objects and then in the **batchExecute()** you will do a flush.

After you've asked how do you access the messages I thought you should have access and I can give access to the messages in the batchExecute() modifying its signature to:

```php
/**
 * @param   AMQPMessage[]  $messages
 *
 * @return   void
 */
public function batchExecute(array $messages);
```

Q: @skafandri what do you think?

skafandri:
I ma not sure if I understand the bulk idea in this context well. What I understood is:
- Get a batch of messages
- Do some work
- Return true to ack them all or false to nack them all
- It should be impossible (as good as we can do) to ack/nack only some messages

devrck:
Yes you've got it.

The problem with this is that if your message generates an exception it will be stuck on your queue forever if you don't have logic to get it off. I can implement a mechanism to know if your messages has been passed along for some time but I don't know if I should do this or the developer.

Whats your opinion?

skafandri:
I think about it to be similar to an SQL transaction

````
start transaction;
insert into order_line(`id`, `product_id`) values (1, 10);
insert into order_line(`id`, `product_id`) values (1, 11);
insert into order_line(`id`, `product_id`) values (1, NULL);
commit;
````
As long as `product_id` is not nullable, all the inserts will fail.
If the bulk operation is not atomic then I wouldn't really need this feature. All I would do is:

````
public function execute(AMQPMessage $msg)
{
           $result = $this->processMessage($msg);
          
           if ($this->reachedBulkCondition()) {
                       $this->bulkAction();
                       $this->resetBulkCondition();
           }
           return $result;
}
````
Where `reachedBulkCondition` can check for the number of processed messages or an elapsed time or both.

Now back to your implementation, if `batchExecute` throws an exception, I think ALL the messages should be nacked. If one of those messages is causing the exception it will effectively get stuck in the queue. But this is the developer's job to handle this situation. The same behavior is present with the single consumer as well.

In other words, I expect `batchExecute` to behave exactly like `execute`. The only difference being the unit of work, `AMQPMessage` for `execute`  and `AMQPMessage[]` for `batchExecute`.

I wouldn't mind creating an `AMQPBulkMessage` which will contain a collection of `AMQPMessage`s to make it clear to the user that he is dealing (nacking/acking) with a collection of messages.

**Comment by devrck** [comment]
Hmm now that I think about my solution is a work around to this `batchExecute(array $messages)` and the existence of execute(AMQPMessage $msg) doesn't seem of any use. I will give another shoot at this and get all the messages in the `batchExecute(array $messages)` part.

**Comment by devrck** [comment]
Fixed CR & implemented another approach to the bulk consumer.

**Comment by devrck** [comment]
Hi guys did anybody got a chance to look over this?

**Comment by devrck** [comment]
A consumer can look like this.

It basically has the same structure as before but now i pass an array of messages to batchExecute (as @skafandri :+1: pointed out that it should work like `execute(AMQPMessage $msg)` and be atomic).


In the first example you can just return one bool value that is the same as saying the same value for each message: 

```php

namespace AppBundle\Service;

use OldSound\RabbitMqBundle\RabbitMq\BatchConsumerInterface;
use PhpAmqpLib\Message\AMQPMessage;

class DevckBasicConsumer implements BatchConsumerInterface
{
    /**
     * @inheritDoc
     */
    public function batchExecute(array $messages)
    {
        echo sprintf('Doing batch execution%s', PHP_EOL);
        foreach ($messages as $message) {
            $this->executeSomeLogicPerMessage($message);
        }

        return true;
    }
}
```

The second example offers you more control on what to do with each individual message but you have to **keep track** of the messages and assign for each **delivery_tag** a value and then return that result back to the batch consumer:

```php

namespace AppBundle\Service;

use OldSound\RabbitMqBundle\RabbitMq\BatchConsumerInterface;
use PhpAmqpLib\Message\AMQPMessage;

class DevckBasicConsumer implements BatchConsumerInterface
{
    /**
     * @inheritDoc
     */
    public function batchExecute(array $messages)
    {
        echo sprintf('Doing batch execution%s', PHP_EOL);
        $result = [];
        /** @var AMQPMessage $message */
        foreach ($messages as $message) {
            $result[(int)$message->delivery_info['delivery_tag']] = $this->executeSomeLogicPerMessage($message);
        }

        return $result;
    }
}
```

@skafandri doing another class `AMQPBulkMessages` is the best choice IMHO for this batch consumer but from what I've investigated (maybe I need more time) you still get one message and then the next.

skafandri:
Looks good for me. I think we should document some edge cases so users won't get confused.

What happens of the batch size is smaller than the prefetch count?
What happens if there are less messages available in the queue than the batch size.
We should also mention that a batch is not fixed. With a batch of 3, if I nack [m1, m2, m3] there is no guarantee (most likely not) that another consumer will receive the same batch, it may receive [m1, m2, m5]

devrck:
First of all this is the configuration for a batch consumer (It the same as above):

```yaml
  # RabbitMQ Configuration
old_sound_rabbit_mq:
    connections:
        default:
            host:     '%rabbit_mq_conn_default_host%'
            port:     '%rabbit_mq_conn_default_port%'
            user:     '%rabbit_mq_conn_default_user%'
            password: '%rabbit_mq_conn_default_password%'
            vhost:    '%rabbit_mq_conn_default_vhost%'
            lazy:     '%rabbit_mq_conn_default_lazy%'
    batch_consumers:
        devck_basic:
            connection:       default
            exchange_options: {name: '%rabbit_mq_exchange_devck%', type: fanout}
            queue_options:    {name: '%rabbit_mq_queue_devck%'}
            callback:         devck.basic
            qos_options:      {prefetch_size: 0, prefetch_count: 2, global: false}
            timeout_wait:     5
            auto_setup_fabric: false
            idle_timeout_exit_code: -2
```

Getting back to your questions:

**_Q_**: What happens of the batch size is smaller than the prefetch count?
**_A_**: The `prefetch_count` is the `batch_size` by my implementation, and I think I have to restrict user to input a value greater or equal to 2 ( I will check and push changes if I haven't yet).

**_Q_**: What happens if there are less messages available in the queue than the batch size?
**_A_**: It will get all the messages it has and after `timeout_wait` of no more messages received on the line you will receive the messages in the `batchExecute()` method. **`timeout_wait` is different from `idle_timeout`**. `timeout_wait` acts like a buffer in case something comes along on the line.

**_Q_**: We should also mention that a batch is not fixed. With a batch of 3, if I nack [m1, m2, m3] there is no guarantee (most likely not) that another consumer will receive the same batch, it may receive [m1, m2, m5] 
**_A_**: I agree there is no guarantee that you will receive the same batch because the dispatching is handled by the RabbitMq server.

Additional situations:

**Prefetch with large messages**:

1. 
   - Queue: 1 messages of 3kb, N messages of 1kb
   - Qos: `prefetch_size`: 3kb , `prefetch_count`: 3
   - In this scenario how does the system behave? (It should receive the first messages of 3kb and send it to the `batchExecute()` because it thats the maximum and it should not stick around `timeout_wait` time for other messages).

2. 
   - Queue: 1 messages of 3kb, N messages of 1kb
   - Qos: `prefetch_size`: 2kb, `prefetch_count`: 3
   - In this scenario I don't see a solution, that 1st message will get stuck on the queue forever correct?

skafandri:
good job, let's add this to README.md so we can review it better to make sure is accurate, helpful, and fits well with the current documentation.

Nightbr:
Any news? It is a great feature !! Awesome works! We really need it!

devrck:
@Nightbr will get back to it to complete the documentation and add revise the code.

christianmeller:
Thank you for this awesome work. This is exactly what we was looking for. 
Could you estimate how long it should take to finish this feature? We really need it as well ;-)

devrck:
@skafandri after a long time I've added the README part.

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+        }
+    }
+
+    public function execute(AMQPMessage $msg)
+    {
+        $this->addMessage($msg);
+
+        $this->maybeStopConsumer();
+
+        if (!is_null($this->getMemoryLimit()) && $this->isRamAlmostOverloaded()) {
```
stloyd: `null !==`

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+    {
+        $processFlags = $this->analyzeProcessFlags($processFlags);
+        foreach ($processFlags as $deliveryTag => $processFlag) {
+            $this->handleProcessFlag($deliveryTag, $processFlag);
+        }
+
+        $this->consumed++;
+        $this->maybeStopConsumer();
+
+        if (!is_null($this->getMemoryLimit()) && $this->isRamAlmostOverloaded()) {
```
stloyd: `null !==`

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+use PhpAmqpLib\Exception\AMQPRuntimeException;
+use PhpAmqpLib\Exception\AMQPTimeoutException;
+use PhpAmqpLib\Message\AMQPMessage;
+
+class BatchConsumer extends BaseAmqp implements DequeuerInterface
+{
+    /**
+     * @var int
+     */
+    protected $consumed = 0;
```
stloyd: If this is not expected to be extended please mark all variables as `private`.

DependencyInjection/Configuration.php
```diff
@@ -13,11 +13,26 @@
...
+    {
+        $this->name = $name;
+    }
+
     public function getConfigTreeBuilder()
     {
         $tree = new TreeBuilder();
 
-        $rootNode = $tree->root('old_sound_rabbit_mq');
+        $rootNode = $tree->root($this->name);
```
stloyd: Why? We don't plan to change the configuration name.

Command/BatchConsumerCommand.php
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * @param   InputInterface  $input
+     */
+    protected function initConsumer(InputInterface $input)
+    {
+        $this->consumer = $this->getContainer()
+            ->get(sprintf($this->getConsumerService(), $input->getArgument('name')));
+
+        if (!is_null($input->getOption('memory-limit')) && ctype_digit((string) $input->getOption('memory-limit')) && $input->getOption('memory-limit') > 0) {
```
stloyd: `null !==`

Command/BatchConsumerCommand.php
```diff
@@ -0,0 +1,113 @@
...
+            // Halt consumer if waiting for a new message from the queue
+            try {
+                $this->consumer->stopConsuming();
+            } catch (AMQPTimeoutException $e) {}
+        }
+    }
+
+    public function restartConsumer()
+    {
+        // TODO: Implement restarting of consumer
```
stloyd: This should be removed or added before merging.

Command/BatchConsumerCommand.php
```diff
@@ -0,0 +1,113 @@
...
+
+class BatchConsumerCommand extends BaseRabbitMqCommand
+{
+    /**
+     * @var BatchConsumer
+     */
+    protected $consumer;
+
+    /**
+     * @return  void
```
stloyd: Such phpdoc is useless IMO.

DependencyInjection/Configuration.php
devrck: Make use of already defined alias.

README.md
```diff
@@ -755,6 +755,100 @@ $ ./app/console_dev rabbitmq:anon-consumer -m 5 -r '#.error' logs_watcher
...
+    {
+        echo sprintf('Doing batch execution%s', PHP_EOL);
+        foreach ($messages as $message) {
+            $this->executeSomeLogicPerMessage($message);
+        }
+
+        // you ack all messages got in batch
+        return true; 
+    }
+}
```
devrck: fixed.

README.md
```diff
@@ -755,6 +755,100 @@ $ ./app/console_dev rabbitmq:anon-consumer -m 5 -r '#.error' logs_watcher
...
+
+```
```
devrck: fixed.

RabbitMq/BatchConsumer.php
devrck: fixed.

RabbitMq/BatchConsumer.php
devrck: fixed.

RabbitMq/BatchConsumer.php
devrck: fixed.

Command/BatchConsumerCommand.php
devrck: fixed.

Command/BatchConsumerCommand.php
devrck: fixed.

devrck:
Waiting review :smile:

DependencyInjection/OldSoundRabbitMqExtension.php
```diff
@@ -358,6 +359,51 @@ protected function loadDynamicConsumers()
...
+                ->addMethodCall('setCallback', array(array(new Reference($consumer['callback']), 'batchExecute')))
+                ->addMethodCall('setExchangeOptions', array($this->normalizeArgumentKeys($consumer['exchange_options'])))
+                ->addMethodCall('setQueueOptions', array($this->normalizeArgumentKeys($consumer['queue_options'])))
+                ->addMethodCall('setQosOptions', array(
+                    $consumer['qos_options']['prefetch_size'],
+                    $consumer['qos_options']['prefetch_count'],
+                    $consumer['qos_options']['global']
+                ))
+            ;
+
```
christianmeller: method call for setIdleTimeoutExitCode (key: idle_timeout_exit_code) is missing

devrck: Fixed.

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+                }
+            } else {
+                $this->batchConsume();
+            }
+
+            if ($this->isCompleteBatch($isConsuming)) {
+                $this->batchConsume();
+            }
+
+            $timeoutWanted = ($isConsuming) ? $this->getTimeoutWait() : $this->getIdleTimeout();
```
stloyd: `(` & `)` are useless here.

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+        }
+    }
+
+    public function batchConsume()
+    {
+        if ($this->batchCounter == 0) {
```
stloyd: `0 === $this->batchCounter`

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,575 @@
...
+            if (!function_exists('pcntl_signal_dispatch')) {
+                throw new \BadFunctionCallException("Function 'pcntl_signal_dispatch' is referenced in the php.ini 'disable_functions' and can't be called.");
+            }
+
+            pcntl_signal_dispatch();
+        }
+
+        if ($this->forceStop) {
+            $this->stopConsuming();
+        } else {
```
stloyd: This whole `} else {` & `return;` is useless.

RabbitMq/BatchConsumer.php
devrck: Fixed.

RabbitMq/BatchConsumer.php
devrck: Implemented.

RabbitMq/BatchConsumer.php
devrck: Fixed.

Command/BatchConsumerCommand.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        if (!AMQP_WITHOUT_SIGNALS && extension_loaded('pcntl')) {
+            if (!function_exists('pcntl_signal')) {
+                throw new \BadFunctionCallException("Function 'pcntl_signal' is referenced in the php.ini 'disable_functions' and can't be called.");
+            }
+
+            pcntl_signal(SIGTERM, array(&$this, 'stopConsumer'));
+            pcntl_signal(SIGINT, array(&$this, 'stopConsumer'));
+//            pcntl_signal(SIGHUP, array(&$this, 'restartConsumer'));
```
stloyd: This looks like unused.

RabbitMq/BatchConsumer.php
```diff
@@ -0,0 +1,573 @@
...
+        } catch (\Error $e) {
+            $this->logger->error($e->getMessage(), array(
+                'amqp' => array(
+                    'queue' => $this->queueOptions['name'],
+                    'message' => $this->messages,
+                    'stacktrace' => $e->getTraceAsString()
+                )
+            ));
+            throw $e;
+        } finally {
```
stloyd: We cannot use `finally` as we support PHP 5.3+

devrck: Sure, fixed.

devrck:
@stloyd what about this PR? What changes do I have to make?

stloyd:
Thank you @devrck !

PR #414, opened 2016-12-08
Adjust Composer autoloader to exclude tests from classmap

stloyd:


skafandri:
@stloyd  This is WIP or can be merged?

PR #415, opened 2016-12-09
Added optional context argument to multiple consumer

tiborb:
Multiple consumer command now accepts a context argument that will be passed to the provider service to change the queue settings


skafandri:
should we update the documentation to mention the new feature?

tiborb:
You are right, i can add it to the docs

Issue #416, opened 2016-12-22
Shift of 5MB not really applied when setting a memory limit ?


JonasHaouzi:
Hello.

The documentation says that when a memory limit is specified (through `-l` or `--memory_limit`), the _consumer will be stopped five MB_.

However, I checked the code, and it looks like those *5 minutes* shift is never applied:
* `BaseConsumerCommand` gets the memory_limit from the argument and pass this limit to the `Consumer` through the method `setMemoryLimit`
* `Consumer::handleProcessMessage` reads messages, and after parsing it try stopping the consumption if the RAM is almost overloaded (through the method `isRamAlmostOverloaded`
* The method `isRamAlmostOverloaded` perform this check: `(memory_get_usage(true) >= ($this->getMemoryLimit() * 1024 * 1024)`

As you can see, no 5 MB shift is applied. 

Please correct me I'm wrong - if not, I can make a little PR to fix that.

Thanks from the @viscaweb team!

skafandri:
You're right, I think it somehow got lost in some merges :)
It would be nice to log the event as well.

PR #417, opened 2016-12-22
Shift of 5MB not really applied when setting a memory limit

JonasHaouzi:
As discussed in the issue #416, here is the PR that apply 5M memory shift to decide whether the RAM is almost overloaded or not.


JonasHaouzi:
Yep, done!

PR #419, opened 2016-12-26
Add idle event and ability to configure exit or not.

makasim:


Event/OnIdleEvent.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * @param boolean $forceStop
+     */
+    public function setForceStop($forceStop)
+    {
+        $this->forceStop = $forceStop;
+    }
+}
```
skafandri: What about adding `$forceStop = true` argument to the constructor and remove the setter altogether?
Events are immutable by nature :)

makasim: @skafandri In my case I'd like to prevent process exit and continue execution. 
That's why I need a setter to change default behaviour

makasim: I working on a cli option `--time-limit="now + 2 minutes"` and to make it work I have to set idle time for wait function but it must exit, just return a control back to php.

makasim: @skafandri And by the way there are a lot of examples even in symfony source code where events are not immutable. and well honestly saying they are not pure events but more like extension points of the bundle logic.

makasim:
Anything I can do to get it merged?

skafandri:
README.md needs to be updated to document the new event

makasim:
@skafandri done.

makasim:
@skafandri is there any ETA for a stable release?

skafandri:
If I have time, this weekend, but most probably next weekend.

Event/OnIdleEvent.php
tPl0ch: @makasim 

> That's why I need a setter to change default behaviour

You could have returned a new instance of the Event with an updated parameters. Immutability is preserved that way.

PR #420, opened 2017-01-03
Bring ProducerInterface uptodate with the Producer

jGielen:


skafandri:
This causes BC break. Will be included in 2.x

coudenysj:
Is there a due date for 2.x? Any tickets were we can help?

mihaileu:
@skafandri why are you seeing this as breaking changes ?

PR #426, opened 2017-01-22
Mention exchange-less usage in the docs.

d-ph:


skafandri:
Can't we summarize this to: default exchange name is **null** and a [link](https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchanges) to the default **direct** exchange in the broker?

d-ph:
Well, I tried to be succinct and this is what I ended up with. I wish someone worded this that way, when I was evaluating this bundle. I.e. that there is an option of configuring the bundle without having one producer per queue, if all I want is a simple queue implementation. I think the copy&paste-ready example is also useful.

I replaced the two link with a single one.

I you really don't like how verbose this paragraph is, then I'll change it.

PR #433, opened 2017-02-14
Added method getCallback() in BaseConsumer

alinilie:


skafandri:
is there a use case to get the callback from the consumer?

alinilie:
My bad, i didn't write about the use case from the first time. Having this method will give you access to the callback from event listener (you already have consumer from method [getConsumer](https://github.com/php-amqplib/RabbitMqBundle/blob/dc96d6ee6d7d10345c037decd4874bcd330dcac8/Event/AMQPEvent.php#L55)).

skafandri:
Ok, so we get the callback which is a `callable`, what can we do with it? Except executing it, which is happening anyway.

alinilie:
I will use the class name to create a metric in Graphite (i have a listener for AMQPEvent::BEFORE_PROCESSING_MESSAGE).

skafandri:
Be aware that it may return any `callable`, so handle it in your code.

PR #435, opened 2017-02-17
Fix isRamAlmostOverloaded function in Consumer

agentsib:


agentsib:
Hello, maybe need change something?

RabbitMq/Consumer.php
```diff
@@ -175,6 +175,6 @@ protected function isRamAlmostOverloaded()
     {
         $memoryManager = new MemoryConsumptionChecker(new NativeMemoryUsageProvider());
 
-        return $memoryManager->isRamAlmostOverloaded($this->getMemoryLimit(), '5M');
+        return $memoryManager->isRamAlmostOverloaded($this->getMemoryLimit().'M', '5M');
```
stloyd: We need to check & adjust if already set memory limit is not in such format.

adm-bome: This check is already done by convertHumanUnitToNumerical() method.

**Just a suggestion:** 
Alter/better documentation in the ReadMe
Its unclear  the `memory-limit` / `-l` option is **not** in MB's by default.

The example `-l 256` is 256 bytes and not 256 mega bytes.
so for reference `-l=256M` will be the correct notation or `-l=268435456`

stloyd:
Thanks @agentsib !

RabbitMq/Consumer.php
agentsib: @adm-bome Ok, I'll do it as soon as I have time. Thanks/

Issue #438, opened 2017-03-15
Specific connection failure in Symfony


needle-project:
In an application with multiple broker connections, if one of the connection is faulty, it is hard to identify it as the only Exception message received is the "Broken pipe or closed connection" event if the issues is just a vhost.

My proposal is to modify the extension class, more specific the "loadConnections" where the factory (in the context of auto-connect) should be surrounded by a try/catch and re-throw the exception decorated with the connection alias.

I will submit a PR for the issues.

PR #439, opened 2017-03-16
Made AMQP Factory aware of the connection alias and detailed exceptio…

needle-project:
Related to issues: https://github.com/php-amqplib/RabbitMqBundle/issues/438

It is not that elegant in context of making the factory aware of something that does not concern it, also, decorating the exception, but it is an useful tool that "patches" the current design.

### DIFF
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -79,7 +79,7 @@ class OldSoundRabbitMqExtension extends Extension
                     : '%old_sound_rabbit_mq.'.$connectionSuffix.'%';
 
             $definition = new Definition('%old_sound_rabbit_mq.connection_factory.class%', array(
-                $classParam, $connection,
+                $classParam, $connection, $key
             ));
             $definition->setPublic(false);
             $factoryName = sprintf('old_sound_rabbit_mq.connection_factory.%s', $key);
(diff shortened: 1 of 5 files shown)


skafandri:
Looks good.
Can you add test cases to `OldSound\RabbitMqBundle\Tests\RabbitMq\AMQPConnectionFactoryTest` and `OldSound\RabbitMqBundle\Tests\DependencyInjection\OldSoundRabbitMqExtensionTest`?

needle-project:
Hello Ilyes, I added tests for the current implementation coverage. I am unaware of the maintainers expectancy so if the tests requires improvements, please be kind to give me some hints and direction about your desired result.

needle-project:
I see that travis failed build from an environment error. Can you force a re-build without new commit?
Thank you

skafandri:
*Broken pipe or closed connection* is thrown from 
- [PhpAmqpLib\Wire\IO\StreamIO::read](https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L207)
- [PhpAmqpLib\Wire\IO\StreamIO::write](https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Wire/IO/StreamIO.php#L269)
- [PhpAmqpLib\Connection\AbstractConnection::wait_frame](https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/AbstractConnection.php#L499)

I don't think a try/catch in the connection factory would catch it because is not thrown on constructor.

PR #446, opened 2017-05-18
Remove support for Symfony <2.7 & allow to install with Symfony 4

stloyd:
Minimal version required for PHP is 5.3.9 to match Symfony 2.7 requirements. Additionally this will allow to install bundle on PHP 7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,65 +1,50 @@
 language: php
 
-matrix:
-  allow_failures:
-    - php: hhvm
-      env: SYMFONY_VERSION=2.3.*
-    - php: hhvm
-      env: SYMFONY_VERSION=2.7.*
-    - php: hhvm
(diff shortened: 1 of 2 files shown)


weaverryan:
Ping @stloyd! Your PR looks good to me, and the tests are passing. 

Could we merge this and create a tag so that people can upgrade to Symfony 4? This is actually one of the most popular Symfony bundles (and 4th most popular that doesn't support Symfony 4 yet).

Thanks!

javiereguiluz:
Changes look OK to me too ... but Travis CI isn't testing Symfony 4 version yet (see https://travis-ci.org/php-amqplib/RabbitMqBundle/builds/233522035) so we may need to configure more jobs to be sure that Symfony 4 is working as expected. Thanks!

weaverryan:
Awesome! Thanks!

Issue #197, opened 2014-09-12
Unable to define multiple producer tags under producers


rrehbein:
Error:
Unrecognized options "0, 1" under "old_sound_rabbit_mq.producers.producer"

Config snippet:

``` xml
<?xml version="1.0" ?>
<container xmlns="http://symfony.com/schema/dic/services"
    xmlns:rabbitmq="http://example.org/schema/dic/old_sound_rabbit_mq"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
<rabbitmq:config>
    <rabbitmq:producers>
        <rabbitmq:producer key="cache_invalidation">
            <rabbitmq:exchange_options name="cache_invalidation" type="fanout" />
        </rabbitmq:producer>
        <rabbitmq:producer key="send_email">
            <rabbitmq:exchange_options name="send_email" type="direct" />
        </rabbitmq:producer>
    </rabbitmq:producers>
</rabbitmq:config>
</container>
```

A work around version

``` xml
<container xmlns="http://symfony.com/schema/dic/services"
    xmlns:rabbitmq="http://example.org/schema/dic/old_sound_rabbit_mq"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

    <rabbitmq:config>
        <rabbitmq:producers key="cache_invalidation">
            <rabbitmq:exchange_options name="cache_invalidation" type="fanout" />
        </rabbitmq:producers>
        <rabbitmq:producers key="send_email">
            <rabbitmq:exchange_options name="send_email" type="direct" />
        </rabbitmq:producers>
    </rabbitmq:config>
</container>
```

PR #451, opened 2017-06-20
Fixing xml configuration for routing-keys

kejwmen:
This is a follow-up to https://github.com/php-amqplib/RabbitMqBundle/pull/210.

Currently setting multiple routing keys using xml configuration:

```xml
<queue-options>
    <name>test-queue</name>
    <routing-keys>
        <routing-key>foo</routing-key>
        <routing-key>bar</routing-key>
    </routing-keys>
</queue-options>
```

Ends with:

```
[Symfony\Component\Config\Definition\Exception\InvalidTypeException]                                                                                   
  Invalid type for path "old_sound_rabbit_mq.consumers.test_consumer.queue_options.routing_keys.routing_key". Expected scalar, but got array. 
```

But works for single routing key, that is confusing:

```xml
<queue-options>
    <name>test-queue</name>
    <routing-keys>
        <routing-key>foo</routing-key>
    </routing-keys>
</queue-options>
```

There is a solution mentioned in https://github.com/php-amqplib/RabbitMqBundle/issues/197

```xml
<queue-options>
    <name>test-queue</name>
    <routing-keys>
        foo
    </routing-keys>
    <routing-keys>
        bar
    </routing-keys>
</queue-options>
```

But it would be nice to use the same valid way that https://github.com/php-amqplib/RabbitMqBundle/pull/210 introduced for routing keys.

After merge, configuration like this:

```xml
<queue-options>
    <name>test-queue</name>
    <routing-key>foo</routing-key>
    <routing-key>bar</routing-key>
</queue-options>
```

Is valid and works as expected.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -402,6 +402,7 @@ class Configuration implements ConfigurationInterface
     protected function addQueueNodeConfiguration(ArrayNodeDefinition $node)
     {
         $node
+            ->fixXmlConfig('routing_key')
             ->children()
                 ->scalarNode('name')->end()
                 ->booleanNode('passive')->defaultFalse()->end()


stloyd:
Thanks @kejwmen !

aosmialowski:
@stloyd can we have a tag, please? For our team, fixing this issue is quite crucial.

Issue #455, opened 2017-07-02
The producer settings are not passed to the class


cirovargas:
Hi

I've passed 3 days on it

Configuration:

```
old_sound_rabbit_mq:
    connections:
        default:
            host:     '%queue_host%'
            port:     '%queue_port%'
            user:     '%queue_user%'
            password: '%queue_password%'
            vhost:    '/'
            lazy:     false
            connection_timeout: 30
            read_write_timeout: 30
            keepalive: false
            heartbeat: 0
            #use_socket: true
    producers:
        search_driver:
            connection:       default
            enable_logger: true
            exchange_options: {name: 'amq.topic', type: topic}
            queue_options:    {routing_keys: ['topic.test']}
    consumers:
        search_driver:
            connection:       default
            enable_logger: true
            exchange_options: {name: 'amq.topic', type: topic}
            queue_options:    {routing_keys: ['topic.test']}
            callback:         pr.consumer.search_driver
```

The call:

`$this->container->get('old_sound_rabbit_mq.search_driver_producer')->publish($message);`

The log:

> [2017-07-01 20:42:03] phpamqplib.DEBUG: AMQP message published {"amqp":{"body":"OOOOI","routingkeys":"","properties":[],"headers":null}} []

```
Producer::publish
print_r($this->queueOptions);
Array
(
    [name] => 
    [passive] => 
    [durable] => 1
    [exclusive] => 
    [auto_delete] => 
    [nowait] => 
    [arguments] => 
    [ticket] => 
    [declare] => 1
    [routing_keys] => Array
        (
            [0] => topic.test
        )
)
```

all the configurations is on the service, but not is passed to de basic_publish php-ampqlib method

cirovargas:
queueBind is called but the message sent not go to the routing key

PR #456, opened 2017-07-02
Possibility of producer/consumer to communicate with MQTT

cirovargas:
Possibility of producer/consumer to communicate with MQTT because MQTT producers/consumers dont recognize queues (work only with topics) and work only on the amq.topic exchange

See:
https://github.com/rabbitmq/rabbitmq-mqtt/issues/141
https://github.com/php-amqplib/RabbitMqBundle/issues/455

### DIFF
--- a/RabbitMq/Producer.php
+++ b/RabbitMq/Producer.php
@@ -53,7 +53,18 @@ class Producer extends BaseAmqp implements ProducerInterface
             $msg->set('application_headers', $headersTable);
         }
 
-        $this->getChannel()->basic_publish($msg, $this->exchangeOptions['name'], (string)$routingKey);
+        if(
+            trim($this->queueOptions['name']) =='' &&
+            $routingKey == '' &&
+            count($this->queueOptions['routing_keys']) > 0
+        ) {
+            foreach($this->queueOptions['routing_keys'] as $configuredRoutingKeys) {
+                $this->getChannel()->basic_publish($msg, $this->exchangeOptions['name'], $configuredRoutingKeys);
+            }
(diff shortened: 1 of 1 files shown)


michaelklishin:
MQTT consumers are backed by queues, so as long as you publish to `amq.topic` (by default) with suitable routing keys — keeping that MQTT and AMQP 0-9-1 topic patterns are converted to each other with some potential for ambiguity because they use different separator — those messages will be enqueued and delivered to MQTT consumers.

[Sender-selected distribution](https://www.rabbitmq.com/sender-selected.html) is a way to work around that but it should not be necessary. `php-amqplib` is perfectly capable of interoperating with MQTT clients as I explained in https://github.com/php-amqplib/php-amqplib/issues/496. At least it is possible to the same extent it is for any other RabbitMQ client.

Unless this bundle makes it impossible to use topic exchanges, I'm not sure what may be missing for MQTT interop.

michaelklishin:
The difference between two code examples in https://github.com/rabbitmq/rabbitmq-mqtt/issues/141 is that one of them does not specify any routing key when publishing. With topic exchanges such messages cannot be routed anywhere. In other words, routing key is mandatory when publishing to a topic exchange.

cirovargas:
Hi @michaelklishin the problem with the bundle is the configured routing_key isnt passed to the publisher on basic_publish method.

the php-amqplib are ok, is possible the integration with mqtt using it.

A possible problem with RabbitMQ (or php-amqplib) is the use of the referenced method queue_bind, the routingKey method parameter dont handle the message to the routing key

Issue #469, opened 2017-09-14
Travis broken


tiborb:
Builds are broken for php 5.3 becasue the php version is not available in the build enviorment.

skafandri:
@tiborb I think is time to bump up PHP version and release a major version, what do you think?

ibrahimgunduz34:
Make sense. I just want to make works. I think 5.4 and 5.5 may removable also, because of even php give support min. 5.6 until dec 2018. Does it make sense ?

skafandri:
I think supporting active versions is the least we can do, so 5.6 seems OK for me.

ramunasd:
We migrated to github actions.

PR #474, opened 2017-09-22
Supported min. php version is upgraded to 5.6

ibrahimgunduz34:
…support on Trusty #469


simPod:
I think the php support might be even increased to 7.1 min

The same as Doctrine project did few months ago http://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

PR #476, opened 2017-09-28
BatchConsumer loop after SIGTERM

nturato:
This pull request solves a bug I discovered on the BatchConsumer. 
When sent a SIGTERM or SIGINT if at least one message of the batch was loaded it would start looping until reaching the maximum function nesting and throwing a fatal error. 

loop chain is:
stopConsumer -> batchConsume -> handleProcessMessages -> maybeStopConsumer -> stopConsumer

this also causes the same batch of messages to be processed multiple times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,14 @@ matrix:
       env: SYMFONY_VERSION=2.8.*
 
   include:
-    - php: 5.3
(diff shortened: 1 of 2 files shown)


stloyd:
Good catch @nturato , thanks for the fix!

PR #484, opened 2017-10-20
Fix AMQPSocketConnection instantiation by the factory

vdechenaux:
`AMQPSocketConnection` and `AMQPStreamConnection` doesn't have the same constructor but the factory ( `AMQPConnectionFactory` ) call them with the same parameters.

https://github.com/php-amqplib/RabbitMqBundle/blob/74b63c0e2807b7b436a6b160849129e2d37cef7b/RabbitMq/AMQPConnectionFactory.php#L63-L78

The 2 constructors:
https://github.com/php-amqplib/php-amqplib/blob/0f90b3d8bc50403458f0eefbcba7d1e2329dd0f6/PhpAmqpLib/Connection/AMQPSocketConnection.php#L33-L36
https://github.com/php-amqplib/php-amqplib/blob/0f90b3d8bc50403458f0eefbcba7d1e2329dd0f6/PhpAmqpLib/Connection/AMQPStreamConnection.php#L18-L22

At this moment, when using `use_socket: true`:
`connection_timeout` is used as `read_timeout`
`read_write_timeout` is used as `keepalive`
`ssl_context` is used as `write_timeout`
`keepalive` is used as `heartbeat`
`heartbeat` is lost (socket construct takes only 13 params, instead 14 for stream)

### DIFF
--- a/RabbitMq/AMQPConnectionFactory.php
+++ b/RabbitMq/AMQPConnectionFactory.php
@@ -4,6 +4,7 @@ namespace OldSound\RabbitMqBundle\RabbitMq;
 
 use OldSound\RabbitMqBundle\Provider\ConnectionParametersProviderInterface;
 use PhpAmqpLib\Connection\AbstractConnection;
+use PhpAmqpLib\Connection\AMQPSocketConnection;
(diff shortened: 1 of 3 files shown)


ramunasd:
Already fixed.

Issue #460, opened 2017-08-08
Publisher confirms (acknowledgements)


mariusbalcytis:
Is it possible to turn on confirms for publishers?

This feature is implemented in both RabbitMQ and php-amqplib for quite some time now:
http://www.rabbitmq.com/blog/2011/02/10/introducing-publisher-confirms/
https://github.com/php-amqplib/php-amqplib/blob/master/demo/amqp_publisher_with_confirms.php

The same issue were opened in #74 but closed by reporter without any further comments.

PR #487, opened 2017-11-05
Add support for publisher confirms for producers

bmichalski:
This PR attempts to solve #460 (and #74 even though it's been closed for no apparent reason by reporter).
It allows to turn confirm mode on for a producer's channel.


UgoTurner:
Hello, it's been a while since that merge request was published. Does someone know why it is on standby ?

bmichalski:
If any change is needed for this PR to be merged, I'd be happy to comply :)
Also it seems to me that this is a pretty important feature: without it there is no way to quickly make sure that the message has been received and persisted by the broker (in order to survive a restart for example).

erild:
Hello, i'm interested in this feature as well.
I manage to test your branch without issue so there doesn't seem to be an issue with the code itself.

Is there any way to know if the maintainers are planning to include it soon ?

umulmrum:
+1 :-)

redshark1802:
would love to have this feature, is there any reason this pr is still open?

redshark1802:
@skafandri @stloyd could you please look at this pr, it's a feature that several users want.

ypereirareis:
Hi, waiting for this feature too !

RabbitMq/PublisherConfirmsProducer.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    protected $publisherConfirmsEnabled = false;
+    protected $publisherConfirmsTimeout = 3;
+
+    public function getPublisherConfirmsTimeout()
+    {
+        return $this->publisherConfirmsTimeout;
+    }
+
+    public function setPublisherConfirmsTimeout($publisherConfirmsTimeout)
```
prgTW: @bmichalski this method could be defined in an interface of some kind. When publisher is of custom class yet still `publisher_confirms: true` is used container extension adds a method call without checking if the method exists. This way we could make the code better if this method is introduced for a purpose (by implementing an interface). WDYT?

PR #488, opened 2017-11-08
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/Tests/Command/BaseCommandTest.php
+++ b/Tests/Command/BaseCommandTest.php
@@ -2,7 +2,9 @@
 
 namespace OldSound\RabbitMqBundle\Tests\Command;
 
-abstract class BaseCommandTest extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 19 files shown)


XWB:
How about bumping PHPUnit dependencies in composer.json to ~4.8.35 || ~5.4.3 ?

gabriel-caruso:
@XWB Did it. Bumped PHPUnit minimum versions.

stloyd:
Merged, thanks @carusogabriel !

PR #495, opened 2017-11-21
Make consumers accessible for public in symfony 3.2.x

koktaildotcom:
In symfony 3.2.x the services are default private within the DI container, making them public prevent the exception: `The "old_sound_rabbit_mq.[....]_consumer" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead. `

see: http://symfony.com/doc/master/service_container/alias_private.html


jojovem:
+1

stloyd:
@koktaildotcom Merged, thanks!

6ecuk:
Hi. Seems in some load methods missing $definition->setPublic(true);
List of methods:
protected function loadMultipleConsumers()
protected function loadDynamicConsumers()
protected function loadBatchConsumers()
protected function loadAnonConsumers()
protected function loadRpcClients()
protected function loadRpcServers()

PR #499, opened 2017-12-01
Fix for Symfony 3.4

syrm:
- service are private by default since Symfony 3.4, we should set it to public
- command should be defined as service since Symfony 3.4


hamaryuginh:
This Pull Request should solve my issue using RabbitMqBundle with Symfony 4.0.
Hope it will be merged soon ;-)

stloyd:
Closing as just merged #500. But thanks for your work!

PR #515, opened 2018-01-03
fix  problem about rabbitmqBundle's rpc, when use symfony4

reywang01:
find the problem like below:
```bash
[vagrant@bogon s4]$ php bin/console  rabbitmq:rpc-server test_server -vvv
2018-01-03T08:00:17+00:00 [error] Error thrown while running command "rabbitmq:rpc-server test_server -vvv". Message: "The "old_sound_rabbit_mq.test_server_server" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead."
2018-01-03T08:00:17+00:00 [debug] Command "rabbitmq:rpc-server test_server -vvv" exited with code "1"
```

In Container.php line 252:
```
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
  The "old_sound_rabbit_mq.test_server_server" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection
   instead.
 ```

### DIFF
--- a/DependencyInjection/OldSoundRabbitMqExtension.php
+++ b/DependencyInjection/OldSoundRabbitMqExtension.php
@@ -104,6 +104,7 @@ class OldSoundRabbitMqExtension extends Extension
                 $definition->setFactoryMethod('createConnection');
             }
             $definition->addTag('old_sound_rabbit_mq.connection');
+            $definition->setPublic(true);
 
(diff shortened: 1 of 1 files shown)


stloyd:
Thanks @reywang01 !

PR #516, opened 2018-01-16
AMPQ Interop migration.

makasim:
The migration to [AMQP Interop](https://github.com/queue-interop/queue-interop#amqp-interop) brings several benefits:

* The producer can be configured to send a message with delay, TTL, and priority.
* Consumer supports PsrProcessor. The code in a processor is not coupled to AMQP, so it could be used with any other [interop compatible transports](https://github.com/queue-interop/queue-interop#implementations). It would be easy to use processors from libraries, like [this](https://github.com/php-enqueue/monolog-queue-handler/blob/master/src/AmqpHandler.php), or [this](https://github.com/liip/LiipImagineBundle/blob/1.0/Async/ResolveCacheProcessor.php), or [this](https://github.com/php-enqueue/enqueue-elastica-bundle/blob/master/Queue/PopulateProcessor.php)
* Stable, semantic OO API. 
* In future it would be possible to use the bundle not only with [php-amqplib](https://github.com/php-amqplib/php-amqplib) but [bunny](https://github.com/jakubkulhan/bunny) or [amqp-ext](https://github.com/pdezwart/php-amqp) as well. So there is no need in creating [same bundle but for amqp ext](https://github.com/M6Web/AmqpBundle) for example.

**The goal is to do migration as backward compatible as possible.**

**At the end, there wont be a dependency on enqueue, only amqp-interop**

### DIFF
--- a/RabbitMq/BaseAmqp.php
+++ b/RabbitMq/BaseAmqp.php
@@ -2,6 +2,11 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+use Enqueue\AmqpTools\RabbitMqDlxDelayStrategy;
+use Interop\Amqp\AmqpContext;
+use Interop\Amqp\AmqpQueue;
+use Interop\Amqp\AmqpTopic;
+use Interop\Amqp\Impl\AmqpBind;
 use OldSound\RabbitMqBundle\Event\AMQPEvent;
 use PhpAmqpLib\Channel\AMQPChannel;
 use PhpAmqpLib\Connection\AbstractConnection;
@@ -118,6 +123,24 @@ abstract class BaseAmqp
     }
 
     /**
+     * @return AmqpContext
+     */
+    public function getContext()
+    {
+        $ch = $this->getChannel();
+
+        // TODO fix it
+        $context = new \Enqueue\AmqpLib\AmqpContext($this->conn);
+        $context->setDelayStrategy(new RabbitMqDlxDelayStrategy());
+        $rp = new \ReflectionProperty($context, 'channel');
+        $rp->setAccessible(true);
+        $rp->setValue($context, $ch);
+        $rp->setAccessible(false);
+
+        return $context;
+    }
+
+    /**
      * @param  AMQPChannel $ch
      *
      * @return void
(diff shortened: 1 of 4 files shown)


makasim:
@stloyd what do you think of this?

makasim:
@stloyd could you please create a 2.x branch?

Evgen14:
Is it planned to continue work on the branch?

makasim:
@Evgen14 I am ready to push this forward, just want to make sure the work is not wasted and would be finally merged.

Evgen14:
@makasim yes, it could be a new version or fork

makasim:
I don’t want work on a fork

Evgen14:
@makasim can I somehow help to promote this PR?

makasim:
Who is in charge of the bundle and making decisions? I am ready to work on this more, just want to be sure it is merged and you maintainers like the idea.

skafandri:
Thank you for the effort. I was checking AMQP interop and it looks promising. I have few comments about DI and tests. I will add them later as I am on mobile now.
Thank you again

makasim:
@skafandri any news? I'd happy to work on this (have some time for that now).

makasim:
@skafandri I tried to keep it as BC as possible. Lots of other things could be improved with breaking changes. 

Though I think we should start with as minimin as possible breaking changes and do bigger things later.

RabbitMq/BaseAmqp.php
```diff
@@ -117,6 +122,24 @@ public function getChannel()
...
     }
 
+    /**
+     * @return AmqpContext
+     */
+    public function getContext()
+    {
+        $ch = $this->getChannel();
+
+        // TODO fix it
```
jderusse: Should be fixed

RabbitMq/BaseAmqp.php
```diff
@@ -117,6 +122,24 @@ public function getChannel()
...
+        $context = new \Enqueue\AmqpLib\AmqpContext($this->conn);
+        $context->setDelayStrategy(new RabbitMqDlxDelayStrategy());
+        $rp = new \ReflectionProperty($context, 'channel');
```
jderusse: Consider avoiding reflection

RabbitMq/BaseAmqp.php
```diff
@@ -117,6 +122,24 @@ public function getChannel()
...
+        $rp->setAccessible(true);
+        $rp->setValue($context, $ch);
+        $rp->setAccessible(false);
```
jderusse: Not needed

RabbitMq/BaseAmqp.php
```diff
@@ -237,15 +294,21 @@ protected function queueDeclare()
...
      */
-    protected function queueBind($queue, $exchange, $routing_key)
+    protected function queueBind($queueName, $exchangeName, $routingKey)
     {
+        $context = $this->getContext();
+
+        $queue = $context->createQueue($queueName);
+        $topic = $context->createTopic($exchangeName);
+
+        // TODO I don't know why we cannot bind to the default exchange. Should be investigated.
```
jderusse: Should be fixed

RabbitMq/Consumer.php
```diff
@@ -2,6 +2,8 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+use Enqueue\AmqpLib\AmqpContext;
+use Interop\Queue\PsrProcessor;
```
jderusse: This is not a PSR. The className is confusing

RabbitMq/Producer.php
```diff
@@ -27,6 +27,54 @@ public function setDeliveryMode($deliveryMode)
         return $this;
     }
 
+    /**
+     * @return null
```
jderusse: What is the point to use a property if the value is always null?

RabbitMq/BaseAmqp.php
```diff
@@ -2,6 +2,11 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+use Enqueue\AmqpTools\RabbitMqDlxDelayStrategy;
+use Interop\Amqp\AmqpContext;
```
jderusse: Package missing in composer's requirements

makasim:
I am trying to keep the migration on to amqp-interop as backward compatible as possible. that's why I added `enqueue/amqp-lib` as a dependency. In the end, everyone could use any interop compatible transports. Hence, no deps on enqueue, only queue-interop\amqp-interop.

PR #517, opened 2018-01-16
Set headers if application_headers key in queueOptions

SriRamanujam:
Sets headers on the queue during bind if `application_headers` is passed in via `queueOptions`.

### DIFF
--- a/RabbitMq/BaseAmqp.php
+++ b/RabbitMq/BaseAmqp.php
@@ -246,7 +246,8 @@ abstract class BaseAmqp
     {
         // queue binding is not permitted on the default exchange
         if ('' !== $exchange) {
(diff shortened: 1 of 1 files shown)


ramunasd:
@SriRamanujam can You elaborate more on this? At first glance all those keywords seems unrelated - application headers, queue options, but applied for exchange? What's the point and what problem are You solving?

SriRamanujam:
It's been so long I can't honestly remember. We ended up going in a different direction with the project that required this fix, so if it's not useful to upstream I think we can go ahead and close this out. I forgot I had even opened this MR until you commented on it, my apologies.

ramunasd:
No worries :)

Issue #496, opened 2017-11-26
How to configure batch consumer to  still working 


alipek:
I want to my batch consumer working permanent. But now when there are less massages in queue as prefech size my consumer is exit after one iteration with timeout. Is there some possibilities to check in loop to message consumer in timeout periods without exit ?

PR #518, opened 2018-01-17
`keep_alive` option for batch consumers #496 

christianmeller:
#496


OClyde:
That's exactly the pull request we need as well! 
Checks don't seem to have failed due to an actual error in the pull request, but that's a general issue, isn't it?

```
rbenv: version '5.3' not installed
The command "phpenv global 5.3" failed and exited with 1 during .
```

Worth to rerun it? Would definitely help a lot to have this in!

stloyd:
Looks good, thanks @christianmeller !

@OClyde I will try to review `.travis.yml` tomorrow, or you also can do it & push PR with fix ;)

PR #519, opened 2018-01-17
Allow to specify that the implementer of ConsumerInterface will handle ACKs

goetas:
In some case (long running consumers) I would like to "ACK" a message and then start the processing.

Did not manage to find a way to disable the ACK feature at rabbitmq level, so for now I'm just sending in the consumer the ACK before the processing as:

```php
public function execute(AMQPMessage $msg)
{
    // manually ACK!
    $msg->delivery_info['channel']->basic_ack($msg->delivery_info['delivery_tag']);

    // do the processing here

    return ConsumerInterface::MSG_ACK_SENT;
   // this instructs the class to not send other ACKs since was already sent manually
}
```


stloyd:
Sounds reasonable too me. Thanks @goetas !

goetas:
:smile:

Issue #457, opened 2017-07-14
"graceful_max_execution" option for rpc server?


maximivanov:
I guess long running rpc server suffers from the same problems as consumers (memory exhaustion, loosing connection to servers). Does it make sense to add `graceful_max_execution` option to rpc server config also?

acasademont:
@maximivanov I think it does, same for multiple or dynamic consumers

easy-dev:
Another issue is that both of them doesn't support signals handling.

PR #524, opened 2018-02-09
Modify RPC server to support signals, execution time , memory limits

easy-dev:
https://github.com/php-amqplib/RabbitMqBundle/issues/457 RPC servers are long-running processes just like any other consumers.
- graceful_max_execution 
- memory limiting
- signals handling


notrix:
👍

PR #525, opened 2018-02-14
Make RPC servers public to avoid errors with Symfony 4

Chrisp1tv:
Maybe I'm wrong, but I think #515 pull request didn't fix completely the problem with Symfony 4, as I'm still getting this error even after the previous fix :

In Container.php line 252:
`The "old_sound_rabbit_mq.XXXX_server" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.`

My commit fixed my problem.


skafandri:
Solved by https://github.com/php-amqplib/RabbitMqBundle/pull/539

PR #526, opened 2018-02-15
Set dynamic consumer public (bugfix for Symfony 4)

NtlBldrv:
Sets dynamic consumers public and fixes this bug that appears in Symfony 4:
"The "old_sound_rabbit_mq.XXX_dynamic" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead."


skafandri:
Solved by https://github.com/php-amqplib/RabbitMqBundle/pull/539

PR #527, opened 2018-02-16
AMQPSocketConnection has a different parameters order, lets consider it

goetas:
`AMQPSocketConnection` has a different parameter ordering than `AMQPStreamConnection`

Currently the example on how to use sockets connection does not work.
You will always get 
```
[PhpAmqpLib\Exception\AMQPIOException]                      
Error reading data. Received 0 instead of expected 7 bytes  
```

To solve is necessary to pass `read_timeout` = 0 or a higher value to the `AMQPSocketConnection` class.

I guess this can be "solved"  by `connection_parameters_provider`, but since sockets looks to be supported natively  by the bundle they should work almost out of the box.

Ive spent days trying to setup socket connections.

### DIFF
--- a/RabbitMq/AMQPConnectionFactory.php
+++ b/RabbitMq/AMQPConnectionFactory.php
@@ -60,22 +60,40 @@ class AMQPConnectionFactory
      */
     public function createConnection()
     {
-        return new $this->class(
-            $this->parameters['host'],
-            $this->parameters['port'],
-            $this->parameters['user'],
-            $this->parameters['password'],
-            $this->parameters['vhost'],
-            false,      // insist
-            'AMQPLAIN', // login_method
-            null,       // login_response
-            'en_US',    // locale
-            $this->parameters['connection_timeout'],
-            $this->parameters['read_write_timeout'],
-            $this->parameters['ssl_context'],
-            $this->parameters['keepalive'],
-            $this->parameters['heartbeat']
-        );
+        if ($this->class == 'PhpAmqpLib\Connection\AMQPSocketConnection' || is_subclass_of($this->class , 'PhpAmqpLib\Connection\AMQPSocketConnection')) {
+            return new $this->class(
+                $this->parameters['host'],
+                $this->parameters['port'],
+                $this->parameters['user'],
+                $this->parameters['password'],
+                $this->parameters['vhost'],
+                false,      // insist
+                'AMQPLAIN', // login_method
+                null,       // login_response
+                'en_US',    // locale
+                isset($this->parameters['read_timeout']) ? $this->parameters['read_timeout'] : $this->parameters['read_write_timeout'],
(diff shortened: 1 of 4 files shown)


goetas:
What surprises me is also that most of the people [here](https://github.com/php-amqplib/php-amqplib/issues/272) as solution for the `Error reading data. Received 0 instead of expected 7 bytes` suggest using streams instead of disabling the read timeout.

Are there different ways to use `AMQPSocketConnection` ?

goetas:
https://github.com/php-amqplib/RabbitMqBundle/pull/527/commits/3a5e688c9029cdbe1455797e378353552f0c3c50 allows to pass generic parameters to constructors allowing to use all other types of connection present in https://github.com/php-amqplib/php-amqplib/tree/master/PhpAmqpLib/Connection

goetas:
Any feedback on this?

skafandri:
Would you mind adding one test for *constructor_args* and one for the `AMQPSocketConnection` case (first if branch)

Tests/RabbitMq/AMQPConnectionFactoryTest.php
```diff
@@ -211,6 +269,28 @@ public function testSSLConnectionParameters()
         ), $instance->constructParams);
     }
 
+    public function testConnectionsParametersProviderWithConstructorArgs()
```
goetas: this tests covers already `constructor_args`. the clas name has to be `OldSound\RabbitMqBundle\Tests\RabbitMq\Fixtures\AMQPConnection`, using any other connection will try to connect to rabbitmq

RabbitMq/AMQPConnectionFactory.php
```diff
@@ -60,22 +60,45 @@ public function __construct(
...
-            $this->parameters['ssl_context'],
-            $this->parameters['keepalive'],
-            $this->parameters['heartbeat']
-        );
+        if (isset($this->parameters['constructor_args']) && is_array($this->parameters['constructor_args'])) {
+            $ref = new \ReflectionClass($this->class);
+            return $ref->newInstanceArgs($this->parameters['constructor_args']);
+        }
+
+        if ($this->class == 'PhpAmqpLib\Connection\AMQPSocketConnection' || is_subclass_of($this->class , 'PhpAmqpLib\Connection\AMQPSocketConnection')) {
```
goetas: do not know really how to test this since will try to connect to rabbtmq

goetas:
Left two comments. is not really clear to me which other tests should be added. Can you provide more info?

RabbitMq/AMQPConnectionFactory.php
skafandri: You are right, I forgot it extends from AbstractConnection

goetas:
:) great, thanks

PR #528, opened 2018-02-21
Changed twig template reference syntax

kejwmen:
Explained the reason here: https://github.com/pixers/PixersDoctrineProfilerBundle/pull/6


stloyd:
Good catch, thanks @kejwmen !

Issue #543, opened 2018-04-04
Symfony 4 support


jamalo:

kejwmen:
Could you explain what exactly does not work?
Support is here for some time https://github.com/php-amqplib/RabbitMqBundle/releases

simivar:
@kejwmen description of package is misleading. It states that this package is only for Symfony **2** appliactions when in reality it works with Symfony 3 and 4 too.

kejwmen:
you're right, thanks you found some time to prepare a PR!

tezrik:
` INFO      [php] User Deprecated: The "OldSound\RabbitMqBundle\Event\AMQPEvent" class extends "Symfony\Component\EventDispatcher\Event" that is deprecated since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead.
[
  "exception" => ErrorException {
    #message: "User Deprecated: The "OldSound\RabbitMqBundle\Event\AMQPEvent" class extends "Symfony\Component\EventDispatcher\Event" that is deprecated since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead."
`

jaztec:
Also relevant at: https://github.com/php-amqplib/RabbitMqBundle/blob/master/RabbitMq/BaseAmqp.php#L271

```
User Deprecated: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Calling the \"Symfony\\Component\\EventDispatcher\\EventDispatcherInterface::dispatch()\" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead. at /path/to/project/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:148)"}
```

roslov:
Any estimations when it goes to a release?

thomaspicquet:
Would be nice to get an update but it doesn't look like this bundle is maintained anymore, no new version since may 2018? :(

simivar:
@roslov @thomaspicquet from what I gathered last year the bundle worked fine. Right now you have official [symfony/messenger](https://github.com/symfony/messenger) component and I would recommend that you use it if you can. [Here are the docs](https://symfony.com/doc/4.1/messenger.html#transports).

[...]

PR #544, opened 2018-04-11
Mark that lib works also with Symfony 3 and 4

simivar:
**Note:** it would be great to update repository description on GitHub as well as it states **Symfony 2**. #543


simivar:
@skafandri could you edit the repository description aswell?

PR #553, opened 2018-06-15
Added new options to consumer

mrygiel:
New options to consumer:
--time-limit
--idle-timeout


mrygiel:
its not needed anymore

PR #559, opened 2018-08-09
Consumer timeout wait

kirkbaly:
Added the ability to set a timeout_wait flag for consumers. This ensures the consumer appropriately terminates when its connection is lost, rather than waiting for an idle timeout or a max execution timeout.


RabbitMq/Consumer.php
```diff
@@ -76,29 +84,35 @@ public function consume($msgAmount)
...
-                    $this->dispatchEvent(OnIdleEvent::NAME, $idleEvent);
+                    $now = time();
 
-                    if ($idleEvent->isForceStop()) {
-                        if (null !== $this->getIdleTimeoutExitCode()) {
-                            return $this->getIdleTimeoutExitCode();
-                        } else {
-                            throw $e;
+                    if ($this->gracefulMaxExecutionDateTime
+                        && $this->gracefulMaxExecutionDateTime <= new \DateTime("@$now")
```
ramunasd, line 101: Why not just use timestamps everywhere? I mean there is no point to use \DateTime here.

ramunasd:
I see there is too much conflicts since last update. Please rebase and reopen as new PR.

RabbitMq/Consumer.php
kirkbaly: Yes, agreed that timestamps would be preferred. I introduced $lastActivityDateTime as \DateTime to match $gracefulMaxExecutionDateTime which was already in place.

kirkbaly:
This PR was opened about 3 years ago back when the repo was not being maintained, so it is perhaps just a little out of date :)

However, it does appear that my changes were introduced by @constantable in 5542098e2cac53cef24dae1ded3772202c5d3411. It looks like they are already in master, so there is no need for this PR or a new one.

Thanks!

Issue #550, opened 2018-06-06
BatchConsumer does not acknowledge gracefulExecutionTime without IdleTimeout


mangalores:
I found some trouble with BatchConsumers ignoring their max execution time.

**problem**
As it seems the code will cancel the consumer when the maxExeution time was reached, but then use idleTimout (=0 by default translating into no timeout) to stay in the wait loop on the cancelled consumerTag forever (as the Server ignores them) without throwing a timeout until the server maybe closes the connection. 

End result can be a bunch of consumers that are running so monitoring tools consider them OK but all of them are cancelled from their queues thus not processing messages.

Another aspect of the current solution is that the maxExecutionTime will not be `graceful_max_execution`, but `graceful_max_execution + idle_timeout ` which at the very least is unintuitive.

**Culprit**
In the BatchConsumer the queue cancellation is called like this:
`$return = $this->getChannel()->basic_cancel($this->getConsumerTag());`

The BaseConsumer uses the following queue cancel call with flags which also unsets the callback
which terminates the loop ends the consumer.

`$return = $this->getChannel()->basic_cancel($this->getConsumerTag(), false, true);`

It should work for the batchConsumer as well but I'm not that deep into the PHP-AmPQ lib to tell if there is a risk of dropping data with the batch routine. Imo it should not as it processes all remaining messages before calling cancel, but someone might want to check that one.

**Steps to reproduce:**
- Configure batch consumer with low max execution time
```
batch_consumers:
        optout_send:
            connection: default
            exchange_options: { name: '##exchange##t', type: topic }
            queue_options:
                name: '##queue##'
            qos_options: { prefetch_size: 0, prefetch_count: 600, global: false}
            timeout_wait:     5
            keep_alive: false
            callback: ##callback##
            graceful_max_execution:
                timeout: 0
```
Consumer will get cancelled out of the RabbitMQ Server consumer list but the script keeps running

Tested with 1.14.4 PHP7.2

.

PR #573, opened 2018-12-01
550-Force exit the batch consumer if gracefullExecutionTimeout

enrico69:
Related to issue #550 
The batch consumer is force to exit if stopConsuming() is called.


mihaileu:
already merged

PR #575, opened 2018-12-13
Support for graceful_max_execution in dynamic consumers

VasekPurchart:
This adds support for `graceful_max_execution` (both timeout and exit code) exactly the same it behaves for basic consumers.

Since the `DynamicConsumer` already extends basic `Consumer` there is no need to change any behavior, just to set up the configuration.


acrobat:
Any chance this can be merged/released? We are in the case that we could use this option in dynamic consumers.

PR #590, opened 2019-06-13
refactor dispatch calls

hjanuschka:
refactors the dispatch calls to get rid of SF4.3 deprecations.


mihaileu:
For using the latest version, which have sy5/4.3 compatibility, please switch to the tag ^2.5.1.

PR #595, opened 2019-08-01
Feature/compatibility with symfony 4.3

nico-vromans:
When using the event dispatcher in Symfony >= 4.3 in combination with the RabbiMQ bundle, you get a deprecation warning:
`php.INFO: User Deprecated: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.`.

To resolve this I've replaced the use statement in **AMQPEvent**: **Symfony\Component\EventDispatcher\Event** with the newer **Symfony\Contracts\EventDispatcher\Event** and changed the order of parameters passed to the **Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch** in **BaseAMQP** (Since Symfony 4.3 the order is flipped: first the event, second (optional) the event name).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
     - php: 7.0
       env: SYMFONY_VERSION=3.0.*
     - php: 7.2
-      env: SYMFONY_VERSION=4.0.*
+      env: SYMFONY_VERSION=4.3.*
     - php: 7.1
       env: DEPENDENCIES=beta
     - php: 7.2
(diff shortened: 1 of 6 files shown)


mihaileu:
Thanks for you PR the issue got fixed here: https://github.com/php-amqplib/RabbitMqBundle/pull/599

Issue #201, opened 2014-10-01
[Feature Request] Events dispatched when grabbing a message, on exception, on ack, etc


adicosoi:
Hello,

I think it will be a really useful feature if events ware dispatched on key point in the consumer.
1. Being able to log all messages consumed
2. Being able to stop exeptions from shooting down your consumer,  grab and ack the message

The consumer is an ongoing command, and the symfony 2 command event don't help because the amqp message can't be accessed from those events

Thank you.

reggin:
Also, very need event onBeforePublish - some values must be in each message (user_id, for example).

videlalvaro:
Aren't these issues related? https://github.com/videlalvaro/RabbitMqBundle/issues/147

Also keep in mind that anything you do during publish will affect performance.

ramunasd:
Implemented. Please see:
https://github.com/php-amqplib/RabbitMqBundle/blob/master/RabbitMq/Consumer.php#L143

Issue #147, opened 2013-12-16
Firing some events when running a command


Taluu:
When using RabbitMQ with Symfony (as this is the goal of this bundle), I think it would be kinda useful to allow to trigger some determined event when receiving a message through a consumer (something like `BEFORE_TREATMENT`, `AFTER_TREATMENT` which would be respectively fired before the callback is called for messages and after it is called, which would get the result returned ?)

I'll try to see if I can make a PR of some sort if the idea could be useful

Keloo:
+1 waiting for the feature.

videlalvaro:
I would consider a PR if someone wants to implement this feature

PR #598, opened 2019-09-18
Dispatch events around publishing

smatyas:
Hi, I saw similar requests in #147 and #201 and I wanted this feature to collect more metrics in my application.

Would you consider merging a PR like this?

Of course, we can work more on the details like event names, places etc., I'd just like to know an early answer if it's a no-go.

### DIFF
--- a/RabbitMq/Producer.php
+++ b/RabbitMq/Producer.php
@@ -2,6 +2,8 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+use OldSound\RabbitMqBundle\Event\AfterPublishingMessageEvent;
+use OldSound\RabbitMqBundle\Event\BeforePublishingMessageEvent;
 use PhpAmqpLib\Message\AMQPMessage;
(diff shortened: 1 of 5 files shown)


smatyas:
@videlalvaro is this something you would consider merging?

mihaileu:
@theodormanolescu what do you think about this ?

RabbitMq/Producer.php
```diff
@@ -53,6 +55,7 @@ public function publish($msgBody, $routingKey = '', $additionalProperties = arra
             $msg->set('application_headers', $headersTable);
         }
 
+        $this->dispatchEvent(BeforePublishingMessageEvent::NAME, new BeforePublishingMessageEvent($this, $msg));
```
theodormanolescu, line 58: Please keep in mind that the signature of the EventDispatcherInterface::dispatch() method has been changed to dispatch($event, string $eventName = null): object since 5.0.0

theodormanolescu (changes requested):

mihaileu:
released in #728

PR #599, opened 2019-10-22
Fix compatibility with Symfony 4.3

talex-de:
Removes deprecation warnings regarding Symfony 4.3:

- The "OldSound\RabbitMqBundle\Event\AMQPEvent" class extends "Symfony\Component\EventDispatcher\Event" that is deprecated since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead.
-  Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.
- Adapt changes to test suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 sudo: false
 
 php:
+  - 7.3
   - 7.2
   - 7.1
   - 7.0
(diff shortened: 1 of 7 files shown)


mihaileu:
For using the latest version, which have sy5/4.3 compatibility, please switch to the tag ^2.5.1.

PR #607, opened 2019-12-06
add php7.3 in travis checking

mihaileu:


mihaileu:
@skafandri @videlalvaro could you please take a look here/

PR #620, opened 2020-09-03
Close RPC client channel on exceptions

notrix:
I found that if consumer calls rpc client and gets exception like time out exception, then channel is left open. This is visible in rabbitmq web panel.


mihaileu:
Thanks for your PR.
Issue got fixed: https://github.com/php-amqplib/RabbitMqBundle/commit/01c7f38cf85bb9bc7cf505bd0ccce38c3f8b684d

PR #621, opened 2020-12-02
Update AMQPConnectionFactory.php

rapaelector:
add option for ampqs like amazone rabbitmq


mihaileu:
already present

https://github.com/php-amqplib/RabbitMqBundle/blame/aaefc33189566bd28d54b2f868d30c7f76a9840e/RabbitMq/AMQPConnectionFactory.php#L1211

PR #631, opened 2021-03-28
Group consumer

sashaaro:
Consume multiple queues by one worker command for reduce complexity and simplify debugging.
A lot of running php commands can consume significant memory size and would be not convinient in development and testing environment which no need parallel executation.

### DIFF
--- a/RabbitMq/GroupConsumer.php
+++ b/RabbitMq/GroupConsumer.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace OldSound\RabbitMqBundle\RabbitMq;
+
+use PhpAmqpLib\Connection\AbstractConnection;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+
+class GroupConsumer extends Consumer
+{
+    /** @var iterable|Consumer[] */
+    protected $consumers = array();
+
+    public function addConsumers(iterable $consumers)
+    {
+        foreach ($consumers as $consumer) {
+            $consumer->setChannel($this->ch);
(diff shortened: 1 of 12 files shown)


RabbitMq/GroupConsumer.php
```diff
@@ -0,0 +1,56 @@
...
+
+class GroupConsumer extends Consumer
+{
+    /** @var iterable|Consumer[] */
+    protected $consumers = array();
+
+    public function addConsumers(iterable $consumers)
+    {
+        foreach ($consumers as $consumer) {
+            $consumer->setChannel($this->ch);
```
pedro-stanaka, line 16: In general, we should cover this class with tests. I think this here is quite important behavior, we should add assertions for it.

RabbitMq/GroupConsumer.php
```diff
@@ -0,0 +1,56 @@
...
+namespace OldSound\RabbitMqBundle\RabbitMq;
+
+use PhpAmqpLib\Connection\AbstractConnection;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+
+class GroupConsumer extends Consumer
+{
+    /** @var iterable|Consumer[] */
+    protected $consumers = array();
+
```
pedro-stanaka, line 12: I would add here:

```php
/**
 * iterable<BaseConsumer>
 */
```

pedro-stanaka (changes requested):
Here are my 2¢. In general, I think you should add some tests to cover the new code and fix some PHPDocs and typos/wording on the README file. I am looking forward on testing this on my end.

sashaaro:
@pedro-stanaka I have not time to finish it. Welcome to contribute it.

PR #632, opened 2021-03-28
[WIP] Experimental 3.0 

sashaaro:
More flexable configuartion, rework bundle yaml config, allow consume multiple queue from different exchanges.
Simplify configure declarations separetly from consumers and producers.
Simplify create and register `Consumer` dynamically
See `README.md`
https://github.com/sashaaro/RabbitMqBundle/blob/experiment/Resources/doc/configuration.md see how we can easy define consumer with many exchanges and callbacks

Execute any consumers by one `rabbitmq:consumer` command only. Add `--no-declare` option
Removed duplicate code.
Removed deprecated features.
Removed `ConsumerInterface` in favor no interface method (inspired Symfony nointefrace controller actions)
Removed anyone consumer classes except common `Consumer` class.
Add kubernetes recipt 

- [ ] Middleware architecture
- [x] ConsumerArgumentResolver
- [ ] Consumer definition by php8 annotation
- [ ] Documentation
- [ ] Examples
- [ ] Test


sashaaro:
@ramunasd is it worth to continue contribute?
I am open for discussion and suggestions
Many things are not done yet but that draft contains implemented features and give understand primary goals for new major release
I would glad to get feedback and help finish it

ramunasd:
The first thing that comes into my head is  - it's a huge change. Second thing - it's very hard to understand what actually is going here, what problem You are solving?
I would suggest to split this change into several smaller and backwards compatible ones. Let's tackle problems one by one instead of trying to fix everything at once ;)

sashaaro:
> I would suggest to split this change into several smaller 

@ramunasd 
I totally rework foundation and configuration with many break changes. Not worth to keep back compatibility, users can stay on v2. It is why I increase major version as v3. 
You can ping me in DM. contact from my github profile

summary:
Remove deprecated code and feature.
Allow consume from multiple queue from different exchanges and handlers. Pls see README.md and example  **important!**
Allow register Consumer dynamically. **important!**
Simplify code and usage, improve event and logging.
Maybe `Consumer definition by php8 class/method annotation`

I have huge work done for present draft of my ideas. Would be glad to get feedback and discuss how we can push it to release.

sashaaro:
https://github.com/sashaaro/RabbitMqBundle/blob/ec0f05e1ee0822e1e6ca6d10dc8a459ed808946c/Resources/doc/configuration.md?plain=1#L81

PR #634, opened 2021-04-09
Do not use internal and deprecated library variable

ramunasd:
Replace internal library variable AMQPChannel->callbacks with AMQPChannel->is_consuming().


mihaileu (approved):
I think you need to increase the minimum requirements for the php-amqplib to 2.9.2

PR #671, opened 2021-11-11
check for type

dvdknaap:
Check for type to prevent never sending an ack message


dvdknaap:
@mihaileu could you merge this asap because more people will have this issue when they upgrade ?

Thanks

mihaileu:
@Eloar  could you take a peek ?

Eloar:
@mihaileu that is a fix for my oversight. I agree it should be merged asap.

mihaileu (approved):

PR #679, opened 2021-12-17
Symfony 5.4 deprecations adressed

Eloar:
Noticed some deprecation messages from RabbitMqBundle when upgraded my project from Symfony 5.3 to Symfony 5.4. Added some more type hinting in Extension class.


mihaileu:
@Eloar  is it working with older sy versions ?

Eloar:
@mihaileu based on tests/checks it does work with lowest version currently supported (Symfony 4.3).

mihaileu:
I think we can merge it after merge this one: https://github.com/php-amqplib/RabbitMqBundle/pull/675 as we will lose that older sy compatibility

DependencyInjection/Configuration.php
```diff
@@ -29,7 +29,7 @@ public function __construct($name)
         $this->name = $name;
     }
 
-    public function getConfigTreeBuilder()
+    public function getConfigTreeBuilder() : TreeBuilder
```
mihaileu, line 32: if there is a space between ) and : we should eliminate it and stick to the PSR12

Eloar: thx for pointing that out. changed all the places I made that mistake and updated commit. It is now ready to be merged if no other issues appear.

phpunit.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
javer, line 1: Why is this file committed when there is the identical one in https://github.com/php-amqplib/RabbitMqBundle/blob/master/phpunit.xml.dist ?

DependencyInjection/OldSoundRabbitMqExtension.php
```diff
@@ -71,7 +72,7 @@ public function load(array $configs, ContainerBuilder $container)
         }
     }
 
-    public function getConfiguration(array $config, ContainerBuilder $container)
+    public function getConfiguration(array $config, ContainerBuilder $container): ?ConfigurationInterface
```
javer, line 75: It's unlikely that this method will return null ever.

javer (changes requested):

DependencyInjection/OldSoundRabbitMqExtension.php
mihaileu: deleted

mihaileu (approved):

PR #684, opened 2021-12-22
lower PHP version requirement

Eloar:
PHP interpreter version requirements lowered to same required by lowest Symfony version currently supported (4.4). Symfony 4.4.35 PHP version requirement is >=7.1.3.

https://github.com/symfony/symfony/blob/v4.4.35/composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "name" : "Alvaro Videla"
     }],
     "require": {
-        "php":                          "^7.4|^8.0",
+        "php":                          "^7.1.3|^8.0",
 
         "symfony/dependency-injection": "^4.4|^5.3|^6.0",
         "symfony/event-dispatcher":     "^4.4|^5.3|^6.0",


mihaileu:
@Eloar thx for the PR, but I think we will not allow older php versions anymore.  I see no reasons for supporting it. 
This is a current statistics regarding php version for our package: https://packagist.org/packages/php-amqplib/rabbitmq-bundle/php-stats#2. There's only 7.7% that is using it with php < 7.4.

They can stil use the package but they will not receive the latest features/fixes until they migrate to php7.4 or bigger. 

Also I was thinking on improving a bit the current code style: https://github.com/php-amqplib/RabbitMqBundle/pull/685

Eloar:
@mihaileu I understand your point. I would like to point two things. First RabbitMqBundle doesn't use currently any feature above PHP7.1. Contributors might have been limited by previous restrictions on PHP version, but there was little PRs with PHP version bump suggested. Second, it is not exactly bundles responsibility to force people into upgrading interpreter if framework doesn't. 

I used to suggest dropping unsupported versions from dependencies constraints. Currently I agree it should have purpose and such change should provide any added value.

I personally use this bundle on 2 projects that are deployed on environment frozen on PHP 7.2.

mihaileu:
we will go this way, php 7.2 is in the past.

PR #685, opened 2021-12-22
Improve code style

mihaileu:


Eloar:
I am always pleased seeing switching to array short syntax.

also `??=` is, if I am correct, first and only PHP 7.4 syntax in this bundle and can be easily kept backward compatible.

PR #689, opened 2022-03-09
Rabbitmq

kuldippujara:


mihaileu:
I do not understand the PR: this bundle already supports sy 5

kuldippujara:
> I do not understand the PR: this bundle already supports sy 5

"Symfony\Component\EventDispatcher\Event" is deprecated since Symfony 4.3, so I replaced it with "Symfony\Contracts\EventDispatcher\Event"

mihaileu:
There are other places where is used the component, could you please search and switch where you can see the Component\EventDispatcher

PR #699, opened 2022-10-25
Add option to register heartbeat sender in consumer

4rthem:
Add `heartbeat-sender` option in consumer command in order to use the [Signal-based Heartbeat](https://github.com/php-amqplib/php-amqplib#signal-based-heartbeat)


mihaileu:
The tests are failing, could you please take a look ?

PR #721, opened 2023-12-18
Add support for Symfony 7

### LINKED ISSUE #720: Symfony 7.0 support

morawskim:
Hello,

This PR should resolve #720.
I mark this PR as draft, because I want to test some other scenarios and versions.

The main problem with upgrade to Symfony7  was with service container. Symfony 7 removed ContainerAwareInterface. 
This interface was [deprecated since 6.4](https://github.com/symfony/dependency-injection/blob/50b32f96f7e8f43b1ec4904df3be04438c0e4c13/ContainerAwareInterface.php#L19)

My solution is to pass ServiceContainer with public services via CompilerPass.

IMHO We should migrate to Service Locator(from my test this works too). There is only issue with backward compatibility. 
The ServiceLocator implements ContainerInterface from Psr namespace not from Symfony.
[The Symfony interface extends from Psr interface](https://github.com/symfony/dependency-injection/blob/50b32f96f7e8f43b1ec4904df3be04438c0e4c13/ContainerInterface.php#L25)
In bundle I did not find any calls to those extra methods, but someone could call one and this break backward compatibility.

@mihaileu if you are interested with migration to ServiceLocator I can prepare another PR.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,11 +16,13 @@ jobs:
         strategy:
             matrix:
                 php-version: ['7.4', '8.2']
-                symfony-version: ['4.4', '5.3', '5.4', '6.0']
+                symfony-version: ['4.4', '5.3', '5.4', '6.4', '7.0']
                 coverage: ['none']
                 exclude:
                     - php-version: '7.4'
-                      symfony-version: '6.0'
+                      symfony-version: '6.4'
+                    - php-version: '7.4'
+                      symfony-version: '7.0'
                 include:
                     - php-version: '8.0'
                       symfony-version: '5.4'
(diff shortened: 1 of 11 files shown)


mihaileu:
thx for PR, we should keep it compatible with symfony 5.4  in order to not maintain 2 branches/tags.

alcohol:
> thx for PR, we should keep it compatible with symfony 5.4 in order to not maintain 2 branches/tags.

Can you elaborate? I'm not sure I understand what you mean with this.

morawskim:
I mark this PR as ready for review.

I tested (with Symfony 7.0 and 6.4):
* command  rabbitmq:stdin-producer
* producer
* command rabbitmq:consumer
* command rabbitmq:multiple-consumer

Everything works.

mihaileu (approved):

mihaileu:
Thx for this PR

PR #728, opened 2024-10-16
Add Producer events

nathanjrobertson:
Currently, Consumers have events which allow users of the bundle to add custom code. This PR adds this support to Producers as well. Really useful for project specific logging, message format validation, etc.

The only real difference between the Consumer and Producer events is the addition of the `routingKey` field in the Events, as `$producer->routingKey` is only set in the case where it is explicitly set, and not in the case where `defaultRoutingKey` is used. This PR currently works around this by providing that extra field, but it might be better to change `Producer.php` (line 67) from:

```php
$real_routingKey = $routingKey !== null ? $routingKey : $this->defaultRoutingKey;
```

to 

```php
if ($routingKey === null) {
    $routingKey = $this->defaultRoutingKey;
}
```

and then removing the extra `routingKey` field from the Events. Happy to take feedback on which way you'd prefer this one go.

My use case for this support is message validation prior to publishing.

### DIFF
--- a/RabbitMq/Producer.php
+++ b/RabbitMq/Producer.php
@@ -2,6 +2,8 @@
 
 namespace OldSound\RabbitMqBundle\RabbitMq;
 
+use OldSound\RabbitMqBundle\Event\AfterProducerPublishMessageEvent;
+use OldSound\RabbitMqBundle\Event\BeforeProducerPublishMessageEvent;
 use PhpAmqpLib\Message\AMQPMessage;
 use PhpAmqpLib\Wire\AMQPTable;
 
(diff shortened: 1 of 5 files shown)


RabbitMq/Producer.php
```diff
@@ -63,6 +65,12 @@ public function publish($msgBody, $routingKey = null, $additionalProperties = []
         }
 
         $real_routingKey = $routingKey !== null ? $routingKey : $this->defaultRoutingKey;
```
mihaileu, line 67: $real_routingKey  == routingKey ?? $this->defaultRoutingKey;

mihaileu (approved):

mihaileu:
thanks foe your pr

mihaileu:
fixed in #729

nathanjrobertson:
Thank you, and apologies for the regression.

PR #731, opened 2024-10-24
 php-amqplib/php-amqplib >= 3.2 required

l-vo:
Due to the usage of PhpAmqpLib\Connection\AMQPConnectionConfig, php-amqplib/php-amqplib >= 3.2 is required.


l-vo:
It's a bug, if php-amqplib/php-amqplib is under 3.2, AMQPConnectionConfig doesn't exist. But actually, AMQPConnectionConfig is only used to retrieve class constants. Using simple strings instead of constants will preserve backward compatibility. WDYT ?

l-vo:
Closed in favor of #732

PR #733, opened 2025-01-03
Remove PHP 8.4.* deprecations

mkoprek:
Just removing some deprecations messages from newest PHP version

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


mkoprek:
Hey @mihaileu; I see that only linters are being run in the pipelines. Shouldn't we add phpunit execution there as well?

PR #740, opened 2025-12-07
Remove deprecations for Symfony 7.4 - XML is deprecated

mkoprek:
Symfony dropped support for configuration in XML


mkoprek:
@mihaileu please take a look here.

PR #741, opened 2025-12-07
Upgrade bundle with current supported version of PHP and Symfony

mkoprek:
Currently supported version of Symfony is 6.4 it time to drop support for older versions of symfony and PHP.


mkoprek:
@mihaileu is it this upload of coverage really required here?

